PR #39, opened 2025-04-28
Replace bit flag helpers with Enums and more

Rurusachi:
- Many bit flags are now Enums
- Change AbilityMap to an InlineArray
- Add AutoAbilityMap
- Add Ability (equipment abilities)


fkelava:
LGTM, thanks for the great work

PR #44, opened 2025-07-19
imageloader API

Rurusachi:


fkelava:
Superseded by dd90f23. Thank you for showing that this is possible, doing the groundwork, and making the original API surface which mostly remains the same in the finished implementation.

PR #45, opened 2025-08-09
Impr: Use `null` over `default` in `FhResourceLoader`

EvelynTSMG:
There's some jankiness going on here with the mod manager stuff... I don't know what's up with that


fkelava:
Merged for FhResourceLoader changes. The May commit that got stuck will be reverted.

fkelava:
Cherry-picked as 3c1d8e7, e3029d7.

PR #47, opened 2025-08-12
Feat: Add introductory mod settings framework

EvelynTSMG:
I'm tired. This isn't _good_, it's just enough to implement the actual save/load IO with. I hope.


fkelava (approved):
Requires fixup, but baseline functional and sufficiently good to iterate on. Test looks fine.

Thank you for the hard work!

PR #52, opened 2025-08-25
FFX-2 support

Rurusachi:
Adds support for FFX-2 + a bunch of FFX-2 data


base/core/FFX2/savedata.cs
```diff
@@ -0,0 +1,309 @@
...
+    public void set_affection_room_gained(int room_id, bool value) { if (room_id < 0x280) affection_room_flags[room_id / 32].set_bit(room_id % 32, value); }
+
+    public uint get_item_count(int item_id) {
+        for (int i = 0; i < 0x44; i++) {
+            if (inventory_ids[i] == item_id) return inventory_counts[i];
+        }
+        return 0;
+    }
+
+    public bool has_key_item(int key_item_id) => key_item_id < 0x80 && important_items[key_item_id / 16].get_bit(key_item_id % 16);
```
fkelava, line 254: we _generally_ do not use expression-bodied methods even for simple expressions; no need to fix this, but just for reference

PR #77, opened 2026-02-19
Add MonStats and ChrPropFlags + fix get_item_count

Rurusachi:
The loop in `get_item_count` should've originally gone to `0x70` instead of `70`. The arrays are actually 256 bytes so I set it to that instead.

I added ChrPropFlags to status.cs since the flags are somewhat similar. The names are based on how they're named in [FFXDataParser](https://github.com/Karifean/FFXDataParser/blob/98a554e8d8515c89bd30cfb1f49ca0e675ff7bcd/src/main/java/model/MonsterStatDataObject.java#L267)

MonStats is the data structure for one of the chunks in monster files (`ffx_ps2/ffx/master/jppc/battle/mon/_monXXX/mXXX.bin`)


core/ffx/Battle/mon_stats.cs
EvelynTSMG: Can you make sure the `always_zero_XX` fields are necessary for `LayoutKind.Sequential` to achieve the correct size for the struct?
If not, they may be padding added by the original compiler, and should not be included.
If yes, these should at least be `private`.

core/ffx/status.cs
EvelynTSMG: Peppy and I have decided `ChrResistFlags` would be a better name.
Sorry for the inconvenience, but could you update this accordingly?

core/ffx/Battle/chr.cs
EvelynTSMG: Peppy and I have decided `special_resistances` would be a better name.
Sorry for the inconvenience, but could you update this accordingly?
You can also remove the `TODO` at that point :3

core/ffx/status.cs
EvelynTSMG: Kari has mentioned `IMMUNITY_ALL_DAMAGE` in actuality prevents only *HP* damage.
Please rename this to `IMMUNITY_HP_DAMAGE` accordingly.
To maintain consistency, please also rename `IMMUNITY_DELAY` to `IMMUNITY_CTB_DAMAGE`.

EvelynTSMG (changes requested):
Needs a few changes with regards to naming and `MonStats`'s fields, but otherwise looks good!

EvelynTSMG (approved):
Looks good to me! Thanks 🩷

PR #80, opened 2026-02-23
Capture-related fields

Rurusachi:
I'm not entirely sure what `battle_type` is. I just named it that based on it disabling most battle rewards when set to a non-zero value


EvelynTSMG (changes requested):
Thank you for contributing to Fahrenheit! We'd appreciate it if you could make some changes before we merge.

EvelynTSMG:
As a sidenote, I'll note I've been prefixing comments explaining game mechanics with "In vanilla, ".
Also, they are most easily justified in events, which may override vanilla code; explaining the unmodified mechanics is therefore useful to any would-be subscribers.

EvelynTSMG (approved):
Looks great! Thank you so much 🩷

PR #86, opened 2026-02-28
prepare Fahrenheit for NuGet packaging, introduce stubbed out mod SDK, retool projects accordingly

fkelava:
once merged, a release should result in the Fahrenheit NuGet package being published, whereupon we will begin testing the mod SDK against it, then publish _it_ as a package in its own right


EvelynTSMG (dismissed):
Thank you! Looks good 🩷

EvelynTSMG (approved):

PR #106, opened 2026-03-11
Implement a few basic events

### LINKED ISSUE #105: Implement basic Events to set a standard

EvelynTSMG:
Resolves #105.

Massively breaking change because it changes `pre_update`/`post_update` to `PreUpdate` and `PostUpdate` events.

### DIFF
--- a/core/runtime/src/events/game_loop.cs
+++ b/core/runtime/src/events/game_loop.cs
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: MIT
 
-namespace Fahrenheit.Modules.Runtime.Events;
+namespace Fahrenheit.Runtime.Events;
 
 [FhLoad(FhGameId.FFX | FhGameId.FFX2 | FhGameId.FFX2LM)]
 public class GameLoopEventsImplModule : FhModule {


core/fh/events/game_loop/main_loop.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Fahrenheit.Events;
+
+/// <summary>
+///     Event arguments for <see cref="GameLoopEvents.PreUpdate">PreUpdate</see>
+///     and <see cref="GameLoopEvents.PostUpdate">PostUpdate</see>.
+/// </summary>
+public readonly ref struct UpdateLoopEventArgs {
+    /// <summary>The time elapsed since the previous frame.</summary>
+    public float delta { get; internal init; }
+}
```
fkelava: ``internal init;`` is correct; the user should not be able to fiddle with it. But the user can still _instantiate_ it (and then do nothing with it), which is not ideal. I suggest event arg types having explicitly ``internal`` ctors.

```cs
internal UpdateLoopEventArgs() { }

// either-or choice
public readonly float delta;
// OR
public float delta { get; internal init; }
```

PR #119, opened 2026-03-20
Add Phyre interop layer, allow game asset loading

### LINKED ISSUE #55: Finalize (or discard) game texture loading
`FhResourceLoader` is capable of loading various format textures from disk or memory successfully, using [DirectXTex](https://github.com/microsoft/DirectXTex) as a base. However, its most appealing and most desired feature is to load _game_ textures at runtime.

This has been blocked on the inability to perform this in a safe way or perform this at all in certain situations, due to the following:
- It has been observed that the game refuses to proceed with loading a texture (or more generally, a Phyre cluster) it has already loaded.
  - If the game is the one to load it first, it is unknown how we are to retrieve it.
  - If _we_ are the ones to load it first, the game will likely hard fault.
- It is unclear how we are to track the lifetime of a cluster. We have no insight into nor control over when the engine chooses to unload the associated clusters.
- It is unclear how we are to track native allocations performed on behalf of callers. This issue is shared with a number of upcoming APIs that will have to resort to storing certain data in unmanaged memory. The caller may have to exercise significant caution, and the entire process including potential failure modes must be exceptionally well documented.

fkelava:
This is a draft PR. Do not merge.

Resolves #55.

We want to enable the use of game assets in ImGui flows. The only permissible way to do so is to load them at runtime, as they must not be shipped with mods. For this to work, the Phyre engine must be called to pry open its custom asset containers.

Phyre types are generally self-describing. That is to say, Phyre classes have 'class descriptors', which contain information about the type such as its name and layout. By inspecting these class descriptors, we can _somewhat_ operate generically over Phyre types.

Where names for classes and/or fields are given, they are sourced from RTTI metadata available in the executables themselves.


src/core/peinterop.cs
```diff
@@ -0,0 +1,603 @@
...
+     *                      LAB_0043e553                                    XREF[1]:     0043e568(j)
+     * 0043e553 8b ce           MOV        ECX,ESI
+     * 0043e555 e8 76 f2        CALL       Phyre::PClassDescriptor::updateBaseOffsets
+     *          ff ff
+     *
+     * See also https://devblogs.microsoft.com/oldnewthing/20040209-00/?p=40713.
+     */
+
+    private static T* abi_adjust_ptr(T* ptr_object) {
+        // Ugly because we cannot switch on the generic type parameter/System.Type properly.
```
EvelynTSMG, line 48: Could do the same as `get_instance_list_filter_for<T>()`?
https://github.com/fahrenheit-crew/fahrenheit/pull/119/changes#diff-f6f123f753d8ff34585c8852f4c16b39e5c6982970b176ad3f796d2c43dcf48eR139

EvelynTSMG: Not that it would necessarily be better, just consistent.

src/core/peinterop.cs
```diff
@@ -0,0 +1,603 @@
...
+    public bool next(out T* item) {
+        item = default;
+
+        if (_m_current == null)
+            return false;
+
+        item = FhPDoubleListIterator<T>.abi_adjust_ptr(_m_current);
+        _m_current = forward
+            ? ((PSimpleDoubleListElement<T>*)_m_current)->next(_m_head)
+            : ((PSimpleDoubleListElement<T>*)_m_current)->prev(_m_head);
```
EvelynTSMG: If `_m_current` is always cast to `PSimpleDoubleListElement<T>*`, why is it defined as a `T*`?

src/core/petypes.cs
```diff
@@ -0,0 +1,600 @@
...
+ * In Ghidra each specialization is presented explicitly.
+ */
+
+[StructLayout(LayoutKind.Sequential, Size = 0x8, Pack = 0x4)]
+internal unsafe struct PSharray<T> where T : unmanaged {
+   /* [fkelava 02/05/26 02:00]
+    * In reality m_u is a union of T and T*,
+    * but this cannot be modeled in a generic.
+    *
+    * The caller must cast to T if required.
```
EvelynTSMG, line 498: In this case, we should probably do
```cs
private T* m_u;

public T* m_u_ptr   => m_u;
public T  m_u_value => (T)m_u; 
```

src/core/resload.cs
```diff
@@ -131,4 +158,46 @@ public bool load_png_from_disk(string file_path, [NotNullWhen(true)] out FhTextu
         texture = null;
         return loader.get_impl(out IFhResourceLoader? impl) && impl.load_texture_from_disk(file_path, FhTextureType.PNG, out texture);
     }
+
+    /// <summary>
+    ///     Attempts to load a 2D Phyre game texture at the given <paramref name="file_path"/> in the game's archives.
+    /// </summary>
+    /// <param name="file_path">The relative file path to the image in the VBF.</param>
+    /// <param name="texture">A <see cref="FhTexture"/> that can be used in ImGui flows.</param>
```
EvelynTSMG: "flows"? Presumably a better wording would be "can be provided to ImGui for rendering".

src/core/resload.cs
```diff
@@ -131,4 +158,46 @@ public bool load_png_from_disk(string file_path, [NotNullWhen(true)] out FhTextu
...
+    public bool load_game_texture_2d(string file_path, [NotNullWhen(true)] out FhTexture? texture) {
+        texture = null;
+        return loader.get_impl(out IFhResourceLoader? impl) && impl.load_game_texture_2d(file_path, out texture);
+    }
+
+    /// <summary>
+    ///     Attempts to load a 3D Phyre game texture at the given <paramref name="file_path"/> in the game's archives.
+    /// </summary>
+    /// <param name="file_path">The relative file path to the image in the VBF.</param>
+    /// <param name="texture">A <see cref="FhTexture"/> that can be used in ImGui flows.</param>
```
EvelynTSMG: "flows"? Presumably a better wording would be "can be provided to ImGui for rendering".

src/core/resload.cs
```diff
@@ -131,4 +158,46 @@ public bool load_png_from_disk(string file_path, [NotNullWhen(true)] out FhTextu
...
+    public bool load_game_texture_3d(string file_path, [NotNullWhen(true)] out FhTexture? texture) {
+        texture = null;
+        return loader.get_impl(out IFhResourceLoader? impl) && impl.load_game_texture_3d(file_path, out texture);
+    }
+
+    /// <summary>
+    ///     Attempts to load a cubemap Phyre game texture at the given <paramref name="file_path"/> in the game's archives.
+    /// </summary>
+    /// <param name="file_path">The relative file path to the image in the VBF.</param>
+    /// <param name="texture">A <see cref="FhTexture"/> that can be used in ImGui flows.</param>
```
EvelynTSMG: "flows"? Presumably a better wording would be "can be provided to ImGui for rendering".

src/core/resload.cs
```diff
@@ -131,4 +158,46 @@ public bool load_png_from_disk(string file_path, [NotNullWhen(true)] out FhTextu
...
+    /// <returns>Whether the operation succeeded and <paramref name="texture"/> can be used.</returns>
+    public bool load_game_texture_cubemap(string file_path, [NotNullWhen(true)] out FhTexture? texture) {
+        texture = null;
+        return loader.get_impl(out IFhResourceLoader? impl) && impl.load_game_texture_cubemap(file_path, out texture);
+    }
+
+    /// <summary>
+    ///     Releases a texture attained through any previous load call.
+    ///     <para/>
+    ///     You may not call this method twice for the same texture, nor with an already otherwise released texture.
```
EvelynTSMG: Can we at least have a separate method (e.g. `request_release_texture`) that is a no-op in these cases?
It could be useful when doing conditional releases, as the caller would not have to care.

src/core/peinterop.cs
EvelynTSMG: If `_m_current` is always cast to `PSimpleDoubleListElement<T>*`, why is it defined as a `T*`?

src/runtime/resloader.cs
```diff
@@ -117,8 +129,204 @@ bool IFhResourceLoader.load_texture_from_disk(string file_path, FhTextureType te
...
+
+    /* [fkelava 01/05/26 18:47]
+     * These functions deal with game textures and call through the Phyre asset load system.
+     * They have a special rule that the 'file path' parameter has to be in the game's canonical form.
+     */
+
+    /// <summary>
+    ///     Attempts to load a 2D Phyre game texture at <paramref name="file_path"/> in the VBF.
+    /// </summary>
+    bool IFhResourceLoader.load_game_texture_2d(string file_path, [NotNullWhen(true)] out FhTexture? texture) {
```
EvelynTSMG: Does `public override bool load_game_texture_2d` not work? Why is this preferred?

src/runtime/resloader.cs
```diff
@@ -117,8 +129,204 @@ bool IFhResourceLoader.load_texture_from_disk(string file_path, FhTextureType te
...
+     * the ImGui module, which executes on the Phyre render thread.
+     *
+     * No `render_imgui` callback will thus run concurrently with this method. This leaves only the case
+     * where a release is attempted from a _different_ thread, and the lock suffices.
+     */
+
+    /// <summary>
+    ///     Flushes the queue of pending resource releases.
+    ///     <para/>
+    ///     ONLY INVOKE THIS METHOD ON THE PHYRE RENDER THREAD.
```
EvelynTSMG:
```cs
    /// </summary>
    /// <remarks>
    ///     Callers should make sure to only call this method on the Phyre rendering thread.
    ///     Calling this method from outside that thread may result in %SOMETHING%.
    /// </remarks>
```

EvelynTSMG (changes requested):
I only have some minor inqueries and requests, good work!

src/runtime/resloader.cs
fkelava: An explicit implementation makes the implementing method private. It is desirable that you go through ``FhApi.Resources`` instead of binding to the implementing module and dry-calling it. In the future I can choose to make substantial alterations in the implementation without breaking the upstream contract.

fkelava: Note also that this is used already in modules using `IFhNativeGraphicsUser.assign_devices` for the same reason. These interfaces are internal/private, and as such their _implementations_ are intended to be internal/private.

EvelynTSMG (approved):

PR #135, opened 2026-05-02
Minor improvements and housekeeping for alpha10

fkelava:
Most of these are self-explanatory and self-contained, except the change to load context logging.

That is included because it is handy in analyzing errors in which a misconfigured/broken build system resulted in loading a separate copy of a library which is meant to be shared; such information cannot be retrieved from crash dumps.


src/core/bootloader.cs
```diff
@@ -216,5 +216,9 @@ internal IEnumerable<FhModuleContext> load_mod(FhManifest manifest) {
 
             yield return new FhModuleContext(module, module_paths);
         }
+
+        FhInternal.Log.Info($"--- load context dump for {manifest.Id} ---");
+        FhInternal.Log.Info(string.Join(',', load_context.Assemblies));
```
EvelynTSMG: Could this message be prepended with 2 spaces or so for a small indent to increase readability?

EvelynTSMG (changes requested):
One minor change I'd like, looks good otherwise!

EvelynTSMG (dismissed):

src/core/bootloader.cs
```diff
@@ -216,5 +216,9 @@ internal IEnumerable<FhModuleContext> load_mod(FhManifest manifest) {
 
             yield return new FhModuleContext(module, module_paths);
         }
+
+        FhInternal.Log.Info($"--- load context dump for {manifest.Id} ---");
+        FhInternal.Log.Info($"  {string.Join('\n', load_context.Assemblies)}");
```
EvelynTSMG: This will print:
```
  assembly1
assembly2
assembly3
```
This should be:
```cs
FhInternal.Log.Info($"  {string.Join("\n  ", load_context.Assemblies)}");
```
or, like I proposed:
```cs
FhInternal.Log.Info($"  {string.Join(",\n  ", load_context.Assemblies)}");
```

EvelynTSMG (changes requested):

EvelynTSMG (approved):

PR #143, opened 2026-05-07
Fix source-gen scripts and move them to `env`-based addressing

### LINKED ISSUE #133: Source-gen scripts are no longer usable in current source tree
The entire contents of `generation` in the source-tree are response files for [ClangSharp](https://github.com/dotnet/ClangSharp), and generate parts of our source directly from game headers remaining in various game versions' data files.

The 'new' canonical way to run Fahrenheit scripts on a game extraction is that you dump all VBF archives to a folder called `env` in the source tree. Generation scripts predate it, and need updating to this standard. DEdit scripts in ``scripts`` already do this.

fkelava:
Closes #133.

The goal is to:
- remove `generation` as a top-level dir, fusing source-gen scripts into either `scripts` or `scripts/sgen`
- fix up slightly broken namespaces that some scripts emit into
- use `env`-relative addressing for all source files, and then make a script to populate an `env` dir from a game install on your system


EvelynTSMG (approved):
I assume the generation is good

PR #144, opened 2026-05-08
Switch to CsWin32 as interop definition provider

fkelava:
Thus far, we've obtained all D3D/Win32/DXGI interop definitions from the [TerraFX.Interop.Windows](https://github.com/terrafx/terrafx.interop.windows) project. 

While there is nothing inherently wrong with it, and the definitions were easy and ergonomic to use, the problem is that TerraFX is a large (~27MB) monolith of a library we have to ship as a result. In typical (AOT, standalone application) use-cases, TerraFX fully supports trimming, which only ships the parts actually used. As a plugin framework, however, we cannot partake in that.

Thus, we're experimentally switching to Microsoft's first-party [CsWin32](https://github.com/microsoft/CsWin32/) source generator. The resulting space savings are approximately 26.5MB. The cost is an increased, but not unreasonably so, complexity of finding and consuming interop types.

Saving this space on disk and in memory would open up room for more useful new features. If testing of this branch passes, we should seriously consider making this the new default.


src/runtime/resloader.cs
```diff
@@ -145,8 +145,16 @@ bool IFhResourceLoader.load_game_texture_2d(FhTexture texture) {
                 return false;
             }
 
+            /* [fkelava 08/05/26 23:41]
+             * TerraFX had a variant that returned HRESULT. For 'reasons' which I genuinely can't fathom, the only
```
EvelynTSMG: "TerraFX, which we previously used, had a (...)"

EvelynTSMG: Also, use `"reasons"` instead of `'reasons'` to avoid ambiguity between ' and \` in some fonts.

EvelynTSMG (changes requested):
Some small notes regarding one comment's wording

EvelynTSMG (approved):

PR #145, opened 2026-05-13
Generate call stubs for all functions

### LINKED ISSUE #90: Find a reliable way to automate Ghidra exports/``FhCall`` generation
We expose game function calls using the ``FhCall`` class. The intent was to export the current state of the Ghidra project and generate the appropriate delegates, because it is simply too error-prone and time-consuming to do otherwise. Users should ideally not have to bring their own delegates when writing Fahrenheit mods.

However, ``FhCall`` is rarely updated because it is surprisingly difficult to automate exports from Ghidra.

A first-party tool to generate C# delegates from a modified Ghidra 'Functions' export, [STEP](https://github.com/fahrenheit-crew/fh-tools-step), exists. It would be better if we could simply use the same ClangSharp tooling we use in ``generation``, but Ghidra's C program export seems to emit largely indigestible output.

This should be improved because ``FhCall`` is an utterly essential convenience of Fahrenheit, and any improvement to it will have a large effect on code cleanliness and the ease of accomplishing various framework tasks.

fkelava:
Lift STEP limitations, causing delegates to be generated for almost all functions.

Move to a mode of handling calls by having a common ``call.g.cs`` and ``call.cs``, and then a pair of each per game. As function calls are analysed and properly annotated, they can be lifted from the corresponding ``call.g.cs`` into ``call.cs``, where they become part of the 'stable' API.

This is a prerequisite for the 'single method handle' approach that has been long envisaged.

Closes #90, in that call stubs will only be generated again if the need arises. Moving forward, methods will be manually annotated as and when necessary.


src/step/main.cs
```diff
@@ -93,14 +93,25 @@ private static void Main(string[] args) {
...
+               !function.Signature.Contains('+');     // ignore descriptively labeled but not authoritatively named functions
+    }
+
+    /// <summary>
+    ///     Determines whether a specific global declaration provided by Ghidra should be interpreted.
+    /// </summary>
+    /// <param name="data_label">The global declaration to be checked.</param>
+    /// <returns>Whether the provided global declaration should be interpreted.</returns>
+    private static bool _should_interpret(FhDataLabelDecl data_label) {
+        return !data_label.Name.Contains('+'); // ignore descriptively labeled but not authoritatively named globals
```
EvelynTSMG, line 114: Did you mean `.StartsWith()`?
Also, where are we generating these to exactly, and why are we ignoring ones we manually named?

src/step/main.cs
```diff
@@ -93,14 +93,25 @@ private static void Main(string[] args) {
...
                !function.Name.Contains("operator") && // ignore operator.new, operator.delete
-               !function.Name.Contains("Unwind@")  && // ignore Unwind@{ADDR} thunks
+               !function.Name.Contains('@')        && // ignore Unwind@{ADDR}, Catch_All@{ADDR} thunks
                !function.Signature.Contains('.')   && // ignore vararg functions
                !function.Signature.Contains(':')   && // ignore anything that even vaguely resembles a C++ namespace
-               !function.Signature.Contains('-');
+               !function.Signature.Contains('`')   && // ignore vector ctors/dtors
+               !function.Signature.Contains('<')   && // ignore template specializations
+               !function.Signature.Contains('-')   &&
+               !function.Signature.Contains('+');     // ignore descriptively labeled but not authoritatively named functions
```
EvelynTSMG, line 105: Did you mean `.StartsWith()`?
Also, why are we ignoring ones we manually named?
Surely, given that they may have no official name (as is oft the case), we should include them with a note specifying the name may change?

src/core/ffx/call.cs
```diff
@@ -0,0 +1,12 @@
...
+/* [fkelava 26/04/26 15:26]
+ * Unlike `call.g.cs`, which contains source-generated delegates with no guarantee of accuracy,
+ * this file contains manually annotated calls with proper Fahrenheit types that are vetted for functionality.
+ *
+ * This file is for calls which are exclusive to FF X and not shared with X-2/LM.
+ */
+
+namespace Fahrenheit.FFX;
+
+public static unsafe partial class FhCall { }
```
EvelynTSMG: Put a blank line between these like so to make future diffs nicer:
```cs
public static unsafe partial class FhCall {

}
```
`{ }` is preferred on types that are supposed to be empty (marker interfaces and the like), but otherwise it just makes the diff of the next commit that adds something worse.

src/core/ffx2/call.cs
```diff
@@ -0,0 +1,12 @@
...
+/* [fkelava 26/04/26 15:26]
+ * Unlike `call.g.cs`, which contains source-generated delegates with no guarantee of accuracy,
+ * this file contains manually annotated calls with proper Fahrenheit types that are vetted for functionality.
+ *
+ * This file is for calls which are exclusive to FF X-2/LM and not shared with X.
+ */
+
+namespace Fahrenheit.FFX2;
+
+public static unsafe partial class FhCall { }
```
EvelynTSMG: Put a blank line between these like so to make future diffs nicer:
```cs
public static unsafe partial class FhCall {

}
```
`{ }` is preferred on types that are supposed to be empty (marker interfaces and the like), but otherwise it just makes the diff of the next commit that adds something worse.

src/core/call.cs
```diff
@@ -0,0 +1,13 @@
...
+ * Unlike `call.g.cs`, which contains source-generated delegates with no guarantee of accuracy,
+ * this file contains manually annotated calls with proper Fahrenheit types that are vetted for functionality.
+ *
+ * This file contains calls which are analogous between both games. Unlike the per-game `call.cs` files,
+ * functions included here are expected to also provide a 'select' helper for automatic .
+ */
+
+namespace Fahrenheit;
+
+public static unsafe partial class FhCall { }
```
EvelynTSMG: Put a blank line between these like so to make future diffs nicer:
```cs
public static unsafe partial class FhCall {

}
```
`{ }` is preferred on types that are supposed to be empty (marker interfaces and the like), but otherwise it just makes the diff of the next commit that adds something worse.

EvelynTSMG (changes requested):

src/step/main.cs
fkelava: They're in the same file, at the bottom. I ignore them until such a time as I/we fix the TODO re: automatically replacing characters which do not constitute valid C# identifiers.

No, I didn't mean ``.StartsWith()``. It works in this case, but not for e.g. template specializations. It's not an optimization I care to make given STEP is already orders of magnitude faster than it needs be.

src/step/main.cs
fkelava: See above thread/comment.

EvelynTSMG (dismissed):
Good work! Thank you :3

EvelynTSMG:
Merge at your convenience after the checks are done.
Edit: It's a draft. I can read.

PR #152, opened 2026-06-02
Rework method handles; generate handles for every function

### LINKED ISSUE #132: Method handle enhancements for alpha10
The following changes are intended for method handles in the next release:
- All parts of ``FhMethodHandle{T}`` which deal with address calculations (including all variant constructors) move into ``struct FhMethodLocation``.
- The current ``FhMethodLocation``, which is used to represent analogous functions at different locations in X and X-2 will either:
  - be renamed ``FhConditionalOffset``, and be one of the variant constructors of ``FhMethodLocation``
  - be removed, and dual ``select``s placed in its current usages instead
- Failure to calculate address/locate method will have nothrow semantics, and failure will propagate/surface at ``hook()`` time
- Method handles now have a single ctor, taking ``FhMethodLocation``

### LINKED ISSUE #153: Hooks sequence in FIFO order, contravening the normal load order
Due to a MinHook implementation detail that went unnoticed until now (https://github.com/TsudaKageyu/minhook/issues/78#issuecomment-485101354), hook sequencing was inverted compared to what we wanted. 

The decided solution:
- Hooks inserted at `init` time are queued for application.
- Hooks are inserted in the proper (LIFO) order after all modules have initialized.
- Hook insertion after `init` is prohibited over a function that already has any.
- Hooks inserted after `init` revert to executing in insertion order.

See #152.

fkelava:
Closes #132, #153. Contributes to like ten more things.

Method handles are now ephemeral ``ref struct``s. State (including complete hook chains) is now tracked and kept alive centrally, and each and every handle generated over a given function has the same view of that state. In other words, instead of 'one handle per function', 'every handle is as good as any other'. Clearer chain syntax is provided to differentiate calling the original (with all hooks) from continuing a hook chain.

``FhCall`` for both games now features method handles for functions which are not explicitly rejected.

Also:
- STEP now generates handles, and separates output into multiple files
- The runtime is completely rewritten to use the new handle style


src/core/handles.cs
```diff
@@ -51,162 +51,228 @@ public void set_impl(T impl) {
...
-        FhInternal.Log.Info($"{hook_fptr.Method.Name}; 0x{addr_target:X8} -> 0x{addr_hook:X8}.");
+/// <summary>
+///     Pairs a hook with its owner. A stack of these constitutes the complete hook chain of a method.
+/// </summary>
+internal sealed record FhHookContext(FhModule owner, Delegate hook);
 
-        FhPInvoke.MH_STATUS rv_create = FhPInvoke.MH_CreateHook(addr_target, addr_hook, &addr_original);
+/// <summary>
+///     Represents a method with signature <typeparamref name="T"/>. You may then invoke or hook it.
+/// </summary>
```
EvelynTSMG: Generally, let's avoid referring to the reader ("you") in documentation. This should be restricted to either Fahrenheit's github wiki or, when appropriate, a modding guide in the FFX/X-2 wiki.

src/core/handles.cs
```diff
@@ -51,162 +51,228 @@ public void set_impl(T impl) {
...
-            return false;
-        }
+    private readonly nint _ptr_target;
 
-        FhPInvoke.MH_STATUS rv_enable = FhPInvoke.MH_EnableHook(addr_target);
+    /// <summary>
+    ///     A pointer to the target function.
+    ///     <para/>
+    ///     Normally, all hooks execute.
+    ///     If <see cref="chain_from(T)"/> is invoked, only the subsequent hooks in the chain run instead.
```
EvelynTSMG: The "if" alread implies an alternative consequence, so the "instead" is tautalogous and should be removed.

src/core/handles.cs
```diff
@@ -51,162 +51,228 @@ public void set_impl(T impl) {
...
+                ? get_fnptr<T>(chain_fnptr)
+                : null;
+        }
+    }
+
+    /// <summary>
+    ///     Inserts the hook located at <paramref name="ptr_chain_link"/> and described by
+    ///     <paramref name="context"/> into the chain of the function at <paramref name="ptr_target"/>.
+    /// </summary>
+    public bool set_fnptr_chain<T>(nint ptr_target, FhHookContext context) where T : Delegate {
```
EvelynTSMG: There is no parameter `ptr_chain_link`. Your IDE ought to warn you as much.

src/runtime/events/game_loop.cs
```diff
@@ -4,33 +4,14 @@ namespace Fahrenheit.Runtime.Events;
...
-    }
+    public GameLoopEventsImplModule() { }
 
     public override bool init(FhModContext mod_context, FileStream global_state_file) {
-        return _h_main_loop.hook()
-            && (_h_jump_to_event?.hook() ?? true);
+        // TODO: Support `PostReturnToTitle` in FFX-2
+
+        return FhCall.h_Sg_MainLoop.hook(this, raise_update_events)
+            && (FhGlobal.game_id is not FhGameId.FFX || FFX.FhCall.h_AtelSetEventJump2.hook(this, handle_warp));
```
EvelynTSMG: For readability, please extract `FhGlobal.game_id is FhGameId.FFX` to a `is_ffx` variable and change this to `&& (!is_ffx || /* hook */)`

src/runtime/gameui.cs
```diff
@@ -38,19 +35,13 @@ internal unsafe delegate nint TOAdpMesFontLXYZClutTypeRGBAChangeFontType(
...
 
     public FhGameUiModule() {
-        FhMethodLocation location_render_game = new(0x4ABCE0, 0x391D00);
-
-        _render_game = new(this, location_render_game, h_render_game);
-
         switch (FhGlobal.game_id) {
             case FhGameId.FFX:
-                _draw_delegate_x = FhUtil.get_fptr<TOMkpCrossExtMesFontLClutTypeRGBA>(0x501700);
+                _draw_delegate_x  = FhUtil.get_fptr<TOMkpCrossExtMesFontLClutTypeRGBA>(0x501700);
```
EvelynTSMG, line 44: Aligning across several other lines feels like a bit much. Ultimately up to you, but still

src/runtime/save_impl.cs
```diff
@@ -28,68 +28,24 @@ internal enum FhSaveExtensionSystemState {
...
-            && _lsm_handle     .try_get_module(out _lsm)
-            && _smm_handle     .try_get_module(out _smm);
+        FhModuleHandle<FhLocalStateModule>  lsm_handle = new(this);
+        FhModuleHandle<FhSaveManagerModule> smm_handle = new(this);
+
+        return FhCall.h_SaveDataManager_debugSave_Internal_6F0650.hook(this, impl_autosave)
+            && FhCall.h_TkMenuJumpToLoadedScene                  .hook(this, impl_copy)
+            && FhCall.h_SaveDataToSave                           .hook(this, signal_enter_save)
+            && FhCall.h_SaveDataToLoad                           .hook(this, signal_enter_load)
+            && (FhGlobal.game_id is not FhGameId.FFX || FFX.FhCall.h_FUN_2EFFF0.hook(this, signal_enter_albd))
```
EvelynTSMG: For readability, please extract `FhGlobal.game_id is FhGameId.FFX` to a `is_ffx` variable and change this to `&& (!is_ffx || /* hook */)`

src/runtime/lifecycle.cs
```diff
@@ -17,22 +17,17 @@ namespace Fahrenheit.Runtime;
...
-        FhMethodLocation location_update_input = new(0x471D10, 0x32CE90);
-
         settings = _settings;
-
-        _update_input = new(this, location_update_input, h_update_input);
     }
 
     public override bool init(FhModContext mod_context, FileStream global_state_file) {
-        return _update_input.hook();
+        return FhCall.h_AtelExec_Internal_871D10.hook(this, handle_input);
```
EvelynTSMG, line 30: Ultimately I'd like `AtelExec_Internal_871D10` to have a custom name but this is fine for now.

EvelynTSMG (changes requested):

fkelava:
Fixes #153.

src/core/handles.cs
```diff
@@ -198,83 +206,137 @@ public T get_fnptr<T>(nint ptr_target) where T : Delegate {
...
+     * This goes directly against _our_ LIFO load order where we want subsequent hooks to take priority.
+     *
+     * One way of proceeding would be to unwind and reapply the entire hook chain, but I could not get it to work.
+     * I assume this is due to https://github.com/TsudaKageyu/minhook/issues/78#issuecomment-485101354.
+     *
+     * Thus we impose the following rules:
+     * - Hooks inserted at `init` time are queued for application.
+     * - Hooks are inserted in the proper order after all modules have initialized.
+     * - Hook insertion after `init` is prohibited over a function that already has any.
+     * - Hooks inserted after `init` revert to executing in insertion order.
```
EvelynTSMG, line 252: The former makes the latter moot, since you cannot hook a function that is already hooked, each post-init-hooked function can only have one hook, meaning there is no ordering to speak of. Therefore, I believe it does not need mentioning.

src/core/handles.cs
```diff
@@ -198,83 +206,137 @@ public T get_fnptr<T>(nint ptr_target) where T : Delegate {
...
+        }
+        catch (Exception e) {
+            FhInternal.Log.Error(e.ToString());
+            return false;
+        }
 
-            // Where do we have to insert the next hook for this chain?
-            nint pTarget = get_fnptr_chain_insert(ptr_target);
-            nint pDetour;
+        unsafe {
```
EvelynTSMG, line 275: Theoretically, it's best practice to include a `//SAFETY: ` comment before every unsafe block (ish) explaining why what you are doing is fine.

Microsoft recommends this in their [article](https://devblogs.microsoft.com/dotnet/improving-csharp-memory-safety/#the-model-in-a-nutshell), but only for members. Coming from [Rust](https://std-dev-guide.rust-lang.org/policy/safety-comments.html), I believe it should apply everywhere.

EvelynTSMG (changes requested):
Some minor nitpicks about comments.

src/core/handles.cs
fkelava: For my own reference later: this limitation is not actually implemented. That is, only functions with hooks at `init` time are protected against further hooking.

EvelynTSMG (approved):

Issue #124, opened 2026-03-25
Make the SDK replace MSBuild properties in the manifest with their values


EvelynTSMG:
The syntax remains to be triaged. I suggest `$(Version)`.

This should support:
- `Id`/`AssemblyName`
- `Name`/`AssemblyTitle`
- `Version`
- `Author`

fkelava:
I forgot to link this to the associated PR. Fixed by #166.

PR #166, opened 2026-06-11
SDK: autogenerate manifest, use portable symbols, create symbol packages

### LINKED ISSUE #163: Debuggers can't retrieve Fahrenheit symbols
See also #162. 

Fahrenheit (both generally and in the SDK) specifies embedded debugging information (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/compiler-options/code-generation#debugtype), such that a portable PDB is contained within the DLL itself.

However, it turns out that such embedded debugging information is not trivially consumable from WinDbg, nor does it seem like Visual Studio itself knows how to obtain symbols unless an exactly matching local copy is present. This is a considerably worse debugging experience than intended, and considerably worse than the fully Source Link enabled (https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink) development experience. 

A better way should be found to supply debug symbols with every Fahrenheit build and release.

fkelava:
Fixes #163 by generating ``.snupkg`` that will be uploaded alongside Fahrenheit NuGet packages.

Source Link now gives 'Valid with Symbol Server', presumably permitting debuggers to retrieve Fahrenheit symbols automatically.

SDK now autogenerates the manifest on behalf of the user based on well-known MSBuild properties.

See also https://github.com/fahrenheit-crew/fh-mods-template/pull/3 for an example of the practical effects.

### DIFF
--- a/src/sdk/Sdk/Sdk.targets
+++ b/src/sdk/Sdk/Sdk.targets
@@ -4,65 +4,30 @@
     <!-- fkelava 30/09/25 23:05
     https://github.com/dotnet/sdk/issues/1366
 
-    When mods reference a Fahrenheit mod, the Framework Library or the Support Library
-    for either game, they must not copy any of those - or their dependencies - to output alongside them.
+    Mods must not include any Fahrenheit core or mod libraries in their build output.
+    The SDK ensures this- and so must the user- by structuring <ProjectReference>s as such:
 
-    This is accomplished by structuring a ProjectReference as such:
-
-    <ProjectReference Include="..\..\core\framework\Fahrenheit.Core.csproj">
+    <ProjectReference Include="...">
         <Private>false</Private>
         <ExcludeAssets>runtime;native</ExcludeAssets>
     </ProjectReference>
 
-    The problem are transitive dependencies. If a mod references a mod which references e.g.
-    the FF X Support Library, MSBuild will fail to respect the chain of ExcludeAssets tags
-    and the Support Library will be emitted alongside the mod, which is invalid and will cause issues.
-
-    This target solves that problem.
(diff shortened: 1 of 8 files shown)


fkelava:
CI fault because CI itself needs amending for implicit publish-on-build

src/sdk/Sdk/Sdk.targets
```diff
@@ -1,68 +1,67 @@
...
 
-    When mods reference a Fahrenheit mod, the Framework Library or the Support Library
-    for either game, they must not copy any of those - or their dependencies - to output alongside them.
-
-    This is accomplished by structuring a ProjectReference as such:
+    Mods must not include any Fahrenheit core or mod libraries in their build output.
+    The SDK ensures this- and so must the user- by structuring <ProjectReference>s as such:
 
-    <ProjectReference Include="..\..\core\framework\Fahrenheit.Core.csproj">
+    <ProjectReference Include="...">
```
EvelynTSMG: Is `"..."` ever a valid path? Does it have a special meaning?

fkelava: It's just a placeholder. It's not valid and has no special meaning. I chose to remove the literal example so as to not make the user believe it is to be copied verbatim, or that it only applies to that specific project. It's a general rule to apply to any project reference to Fahrenheit core (handled by SDK) or another mod (handled by user).

EvelynTSMG: In that case, it should be replaced with "PATH_TO_PROJECT_HERE" or similar

EvelynTSMG (changes requested):

EvelynTSMG (approved):

fkelava:
Fixes #124. Forgot to note that.

PR #167, opened 2026-06-11
Disable automatic code analysis for the main project

### LINKED ISSUE #164: Run-as-you-type analysis mode impairs the development experience
In modern .NET IDEs (Rider/VS), Roslyn analyzers run as you type, and solution-wide analysis is enabled by default. This allows errors to surface before build-time, aiding the development experience. However, in sufficiently modern VS (>18.2), a solution of Fahrenheit's size causes patently excessive memory usage. We should strongly consider disabling both options by default.

See https://developercommunity.visualstudio.com/t/DevHubexe-using-13GB-of-memory-and-util/11041281,

fkelava:
Closes #164.

This PR is drafted as one of the potential ways we can address the fact that in both Rider and VS, real-time code analysis is both time- and memory-consuming with a solution of the sheer size of Fahrenheit. Furthermore, VS has recently had some issues of its own with excessive memory consumption that is a significant drain on performance. 

We should therefore _consider_ whether disabling automatic code analysis (and thus forcing devs interacting with this repo to run it manually at their leisure) provides a compelling benefit. If not, then the PR should be rejected.

### DIFF
--- a/src/core/Fahrenheit.csproj
+++ b/src/core/Fahrenheit.csproj
@@ -47,6 +47,12 @@
         <Product>Fahrenheit Mod Framework</Product>
     </PropertyGroup>
 
+    <!-- CODE ANALYSIS CONTROLS -->
+    <PropertyGroup>
+        <RunAnalyzersDuringBuild>false</RunAnalyzersDuringBuild>
+        <RunAnalyzersDuringLiveAnalysis>false</RunAnalyzersDuringLiveAnalysis>
+    </PropertyGroup>
+
     <!-- SHARED FILE LINKS -->
     <ItemGroup>
         <None Include="$(SolutionDir)\.editorconfig" Link=".editorconfig" />


EvelynTSMG:
Performance results for me (Intel i7 7700k, JetBrains Rider):

Build times without this change:
- 1st build: 3m 57s
- Subsequent builds: 4m 02s

Build times with this change:
- 1st build: 3m 36s
- Subsequent builds: 3m 39s

`call_*.g.cs` file load and navigation times have improved a noticable amount with this change.
I do not know how much it used before, but Rider is currently using ~6.5 GB of RAM. This is, of course, absurd, but I suppose such is just the nature of IDEs these days.

EvelynTSMG (approved):

fkelava:
The amount of time it takes to do just about anything in ``call_*.cs`` merits this. We can revisit with more data over time.

PR #168, opened 2026-06-11
Add limited first-chance and unhandled exception handling

### LINKED ISSUE #162: Debugging on non-Windows platforms is severely impaired by platform constraints
Fahrenheit, like the game itself, supports non-Windows platforms only incidentally. 

When issues arise under WINE/Proton, obtaining functional debugging information is nearly impossible. Not only will Win32 DLLs fail to symbolicate, managed exception information can also be completely trashed in resulting core dumps, leaving us functionally unable to address the issue.

Stage0 is normally meant to show managed exception information and an exit code, but users often have it set up to vanish when the game does, leaving _that_  off the table as well. We need a way to surface fault information more reliably.

fkelava:
Fixes #162.

This only serves to log normal managed exceptions. That is, "hard faults" are not interceptible.
See https://learn.microsoft.com/en-us/dotnet/api/system.appdomain.firstchanceexception?view=net-10.0#remarks:
> This event is not raised for exceptions that indicate corruption of process state, such as access violations {...}

In .NET Framework an attribute existed to override this (https://learn.microsoft.com/en-us/dotnet/api/system.runtime.exceptionservices.handleprocesscorruptedstateexceptionsattribute?view=net-10.0), but .NET Core has no such provision:
> Even though this attribute exists in .NET Core, since the recovery from corrupted process state exceptions is not supported, this attribute is ignored. The CLR doesn't deliver corrupted process state exceptions to the managed code.

### DIFF
--- a/src/core/log.cs
+++ b/src/core/log.cs
@@ -32,60 +32,65 @@ public class FhLogger {
         _file    = new TextWriterTraceListener(File.Open(log_path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.ReadWrite));
     }
 
-    internal void Log(FhLogLevel level,
-                      string     msg) {
-        if (level < MinLevel) return;
-
-        _console.WriteLine(msg);
-        _file   .WriteLine(msg);
+    /// <summary>
+    ///     Logs a given <paramref name="message"/> unconditionally,
+    ///     without timestamp or caller information.
+    /// </summary>
+    internal void Log(string message) {
+        _console.WriteLine(message);
+        _file   .WriteLine(message);
         _file   .Flush();
     }
 
+    /// <summary>
+    ///     Logs a given <paramref name="message"/> at a given <paramref name="level"/>,
+    ///     prepending a timestamp and caller information.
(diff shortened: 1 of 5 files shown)


src/core/log.cs
```diff
@@ -32,13 +32,22 @@ public FhLogger(string log_file_name) {
         _file    = new TextWriterTraceListener(File.Open(log_path, FileMode.OpenOrCreate, FileAccess.ReadWrite, FileShare.ReadWrite));
     }
 
+    internal void Log(FhLogLevel level,
+                      string     msg) {
```
EvelynTSMG: Firstly, put these on one line. There is no reason to chop them like the public Log method.
Secondly, does this not cause a collision if we tried to use the public method internally?

EvelynTSMG (changes requested):

src/core/log.cs
fkelava: it does; I've amended the logging methods accordingly

EvelynTSMG (approved):

PR #169, opened 2026-06-12
Add native exception handler, write customized core dumps

fkelava:
Contributes to #162.

#168 added limited exception logging on the managed side. This is, by design, limited to non-fatal managed exceptions. However, we have a vested interest in capturing _all_ exceptions, including such things as AVs.

For this, we have to resort to native exception handling. We override the game's own unhandled exception filter (https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-unhandledexceptionfilter, https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-setunhandledexceptionfilter) to substitute our own. This has a dual effect:
- we can control the exact dumping behavior
- we avoid a case where the games refuse to dump core

The resulting dumps _seem_ fine, but Visual Studio sometimes has issues reconstructing mixed frames. There is also the open question of whether this will improve the dump debugging experience under WINE/Proton. WinDbg on Windows- the primary tool for heavyweight debugging- accepts these dumps as they are. We should move forward with testing.

The correctness should be vetted in an actual project on a variety of systems because unlike the managed side, things can go _badly_ wrong here.

### DIFF
--- a/src/stage1/src/main.cpp
+++ b/src/stage1/src/main.cpp
@@ -134,7 +134,7 @@ stage1_eh_filter_dump(
 
         case IncludeThreadCallback: {
             // Exclude the thread which writes the minidump.
-            return CallbackInput->Thread.ThreadId != g_eh_thread_handler_id;
+            return CallbackInput->IncludeThread.ThreadId != g_eh_thread_handler_id;
         } break;
     }
 
--- a/src/stage1/src/main.cpp
+++ b/src/stage1/src/main.cpp
@@ -2,92 +2,224 @@
  * Uses code from the .NET NativeHost sample, used under the MIT license.
  *
  * See THIRD-PARTY-NOTICES.
+ *
+ * For the HostFXR bits, see https://github.com/dotnet/samples/blob/main/core/hosting/src/NativeHost/nativehost.cpp.
+ * For the exception handling, see:
+ * - http://code.aaronballman.com/minidumper/MiniDump.cpp
+ * - https://github.com/folgerwang/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Private/Windows/WindowsPlatformCrashContext.cpp
  */
 
 #include "fhstage1.h"
 
-typedef void (CORECLR_DELEGATE_CALLTYPE* fh_ldr_managed_init)();
+// Function pointer to managed delegate with our own signature
+typedef void (CORECLR_DELEGATE_CALLTYPE* fh_init)();
 
 using string_t = std::basic_string<char_t>;
-using main_t   = int(*)(void);
+using main_fn  = int(*)(void);
+using eh_fn    = LONG(*)(EXCEPTION_POINTERS*);
 
-main_t pMainOriginal = nullptr; // Original program entrypoint.
-main_t pMainTarget   = nullptr; // Detour of program entrypoint.
+// Globals to hold original and detour addresses of program entrypoint and SEH filter
+main_fn g_fnptr_main_original = nullptr;
+main_fn g_fnptr_main_target   = nullptr;
+eh_fn   g_fnptr_eh_original   = nullptr;
+eh_fn   g_fnptr_eh_target     = nullptr;
 
-namespace {
-    // Globals to hold hostfxr exports
-    hostfxr_initialize_for_runtime_config_fn init_fptr;
-    hostfxr_get_runtime_delegate_fn          get_delegate_fptr;
-    hostfxr_close_fn                         close_fptr;
+// Globals for EH override
+DWORD               g_eh_thread_faulting_id;
+DWORD               g_eh_thread_handler_id;
+HANDLE              g_eh_thread_handler;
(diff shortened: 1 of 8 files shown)


fkelava:
For why the original games refuse to dump core in certain circumstances, see:
- ``FFX.exe+226AB3``
- ``FFX-2.exe+6B6363``

fkelava:
It seems known that the VS debugger has... issues... with mixed frames. See https://learn.microsoft.com/en-us/visualstudio/debugger/mixed-code-and-missing-information-in-the-call-stack-window?view=visualstudio.

fkelava:
Reverting to draft in light of the fact certain exception types are completely swallowed by the managed side. See https://github.com/dotnet/runtime/issues/101560.

src/stage1/src/main.cpp
```diff
@@ -2,92 +2,224 @@
...
-        assert(f != nullptr);
-        return f;
+
+    return TRUE;
+}
+
+// Writes a customized core dump.
+static DWORD CALLBACK stage1_eh_create_dump(LPVOID lpThreadParameter) {
+    HANDLE hFile = CreateFileW(
+        L"fahrenheit.dmp",
```
EvelynTSMG: Most players won't have file extensions enabled, so producing a file called `fahrenheit` next to the folder of the same name makes it harder to communicate what they should send over. Let's call this `fahrenheitCoreDump.dmp` instead.

EvelynTSMG: To be clear, even if this would get created in the bin folder, it would still be better to name it something else.

EvelynTSMG (changes requested):

EvelynTSMG (dismissed):

goaaats:
Hey! I saw this change on the runtime issue, we have a somewhat seasoned custom crash handler that logs mixed-mode callstacks with managed and unmanaged symbol names, produces dumps that can be properly debugged in VS, catches CLR fast-fails and works on Windows and Wine. I see that your project is MIT but I'd be happy to grant you an exception if you want to adapt it until MS makes an API that works. It's certainly a bit of a mess, but maybe you can make use of it.

https://github.com/goatcorp/Dalamud/blob/master/DalamudCrashHandler/DalamudCrashHandler.cpp
https://github.com/goatcorp/Dalamud/blob/master/Dalamud.Boot/veh.cpp
https://github.com/goatcorp/Dalamud/blob/master/Dalamud.Boot/dllmain.cpp#L448

Sorry if this is not appreciated :)

fkelava:
It's very much appreciated, thank you. 😄  

If I can wrap my head around it (and how to fit it into our loader), I'll contact you regarding an exception.

fkelava:
Not mergable until we've relicensed, as to permit the utilization of the one method taken from Dalamud VEH code.

EvelynTSMG:
We have now relicensed, so this is mergable.

fkelava:
It's worth another look because the latest commit has changed some things.

src/stage1/src/main.cpp
```diff
@@ -36,7 +38,8 @@ hostfxr_get_runtime_delegate_fn          g_fnptr_hostfxr_get_delegate;
 hostfxr_close_fn                         g_fnptr_hostfxr_close;
 
 // Using the nethost library, discover the location of hostfxr and get exports
-static bool load_hostfxr() {
+static bool
+load_hostfxr() {
```
EvelynTSMG: What? Why did the formatting change here? It was fine.

EvelynTSMG: Same for other functions here.

src/stage1/src/main.cpp
```diff
@@ -389,10 +464,11 @@ static int stage1_main(void) {
...
 
-BOOL APIENTRY DllMain( HMODULE hinstDLL,
-                       DWORD   fdwReason,
-                       LPVOID  lpvReserved
-                     ) {
+BOOL APIENTRY
+DllMain( HMODULE hinstDLL,
+         DWORD   fdwReason,
+         LPVOID  lpvReserved
+         ) {
```
EvelynTSMG: I think the formatting should be:
```suggestion
BOOL APIENTRY DllMain(
    HMODULE hinstDLL,
    DWORD   fdwReason,
    LPVOID  lpvReserved
) {
```

Also, it isn't new, but why are we using Hungarian notation? I get that this is C++ and that's what the Windows API uses but let's not fall to tradition without reason.
```suggestion
BOOL APIENTRY DllMain(
    HMODULE dll,
    DWORD   reason,
    LPVOID  reserved
) {
```

EvelynTSMG (changes requested):
I trust Dalamud's code to be correct, I skimmed the rest.

What's up with the formatting?

src/stage1/src/main.cpp
fkelava: Basic idea was to do as MSDN does. Amended either way.

EvelynTSMG (approved):

PR #175, opened 2026-06-15
Save and load mod settings

### LINKED ISSUE #54: Finalize the 'mod settings' API
While the 'mod settings' feature was tentatively merged in PR #47, it is not implemented/wired up fully today.

This issue lists the outstanding tasks for that to be done:

- [ ] Actually wire up the save/load callbacks to emit to ``mod.config.json``
- [ ] Clean up and thoroughly comment each possible component/UI element

### LINKED ISSUE #123: State hash evolution for RC1
When we introduced the concept of 'local' (i.e. per-save) state and 'separate' (i.e. per-mod-list) saves, the backing concept became the _state hash_, a [SHA256 hash of a concatenated string of mod IDs](https://github.com/fahrenheit-crew/fahrenheit/blob/9fb4aea6bc7ac8d9d57f94de23a761ed1c0f2057/core/fh/bootloader.cs#L44) which set the [manifest flag ``SEPARATE_SAVES``](https://github.com/fahrenheit-crew/fahrenheit/blob/9fb4aea6bc7ac8d9d57f94de23a761ed1c0f2057/core/fh/module.cs#L49). Effectively, a state hash uniquely identifies a unique load order of such mods.

This is functional, but it has the downside that full, non-truncated SHA256 hashes result in paths in the install directory that exceed ``MAX_PATH`` for a typical install location. Users thus cannot browse state on disk using Windows Explorer- while not breaking, because file operations in .NET (and thus in both Fh and the mod manager) ignore the ``MAX_PATH`` limit, users should still be able to do so.

We should find a better way to do this. If the manifest ID is replaced by the assembly name due to #122, we will have to make changes to _what_ we are hashing anyway, so it's a good opportunity to evaluate:

- Does a more suitable (quicker, equally resistant to collision/second preimage) hash algorithm for this exist?
- How would truncating the hash affect properties such as collision/second preimage resistance? What is a reasonable length for a modlist hash?
- How do other mod managers, mod systems, or games compute modlist hashes? We can look at:
  - Hearts of Iron IV
  - Total War: Warhammer III (and its TWMM mod manager)
  - {...}
- Could we change _what_ we are hashing? Is there a better input then concatenated mod IDs/``AssemblyName``s?

fkelava:
This is a work in progress.

Closes #54, closes #123.

Wire up the 'mod settings' API. To facilitate this, hashing is being taken out of `FhEnvironment` and into an internal `FhHasher`, which provides an amended state hash (renamed the 'save set' hash), and a general modlist hash. The hash tentatively uses BLAKE3, and is shortened to match what other games provide.


src/core/pathmgr.cs
```diff
@@ -27,6 +28,7 @@ public sealed record FhModPaths(
 /// </summary>
 internal sealed class FhFinder {
     private const string _dirname_bin   = "bin";
+    private const string _dirname_cfg   = "cfg";
```
EvelynTSMG: While I can appreciate the brevity, I believe naming this folder `config` would be better for explaining to others where the settings are stored in the future, as not everyone would be able to figure out that "cfg" stands for "config".

EvelynTSMG (changes requested):
So far so good!

src/core/settings.cs
```diff
@@ -0,0 +1,249 @@
...
+        foreach (FhSetting setting in settings) {
+            setting.save(writer);
+        }
+    }
+
+    /// <inheritdoc cref="FhSetting.load" />
+    internal override void load(Utf8JsonReader reader) {
+        foreach (FhSetting setting in settings) {
+            setting.load(reader);
+        }
```
EvelynTSMG, line 160: This fails to load the (properly saved) `{id}.collapsed` setting.

src/core/settings.cs
```diff
@@ -0,0 +1,249 @@
...
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit;
+
+/// <summary>
+///     An internal class of setting-related utilities and constants.
+/// </summary>
+internal sealed class FhSettings {
```
EvelynTSMG: I am of the solid belief that short single-line documentation comments SHOULD be folded onto one line along with the tags:
```suggestion
/// <summary>Internal API including setting-related utilities and constants.</summary>
internal sealed class FhSettings {
```
\
(Additionally, using "class" in the doc comment for a class is pointless repetition. You could argue "**Internal** API" on an internal class is as well, but there it serves to deny the usual assumption of an API being for public consumption)

src/core/settings.cs
```diff
@@ -0,0 +1,249 @@
...
+        writer.WritePropertyName(id);
+        JsonSerializer.Serialize<T>(writer, _value, FhUtil.InternalJsonOpts);
+    }
+
+    internal sealed override void load(Utf8JsonReader reader) {
+        /* [fkelava 22/06/26 17:50]
+         * A `Utf8JsonReader` is forward-only. Since `try_find_key_and_deserialize` will read
+         * as far ahead as necessary to find the key, we need to copy the reader.
+         *
+         * The same is done by S.T.J when necessary. See dotnet/runtime,
```
EvelynTSMG: I'll be honest I have no idea what "S.T.J" means, I assume the json serializer C# provides? If so please specify the full name at least once (the first time) like so: "Super Transgender Jaguar (S.T.J.)" for people like me chancing upon this in the future.

EvelynTSMG (changes requested):
Some small notes here and there

src/core/settings.cs
```diff
@@ -139,9 +130,7 @@ public FhSettingsCategory(string id, FhSetting[] settings) : base(id) {
         update_ids();
     }
 
-    /// <summary>
-    ///     Prepends the category's ID to all subordinate settings.
-    /// </summary>
+    /// <summary>Prepends the category's ID to all subordinate settings.</summary>
```
EvelynTSMG, line 133: Id is a word, not an acronym, therefore it should be lowercase.
```suggestion
    /// <summary>Prepends the category's id to all subordinate settings.</summary>
```

src/core/settings.cs
```diff
@@ -161,6 +150,11 @@ internal override void save(Utf8JsonWriter writer) {
 
     /// <inheritdoc cref="FhSetting.load" />
     internal override void load(Utf8JsonReader reader) {
+        Utf8JsonReader copy = reader;
+        if (copy.try_find_key_and_deserialize($"{id}.collapsed", out bool is_collapsed)) {
+            collapsed = is_collapsed;
+        }
```
EvelynTSMG, line 156: Could this be simplified? I am not sure.
```suggestion
        copy.try_find_key_and_deserialize($"{id}.collapsed", out collapsed);
```

src/core/settings.cs
EvelynTSMG: I am of the solid belief that short single-line documentation comments SHOULD be folded onto one line along with the tags:
```cs
/// <summary>A text input setting, with flags. Callbacks are not currently supported.</summary>
```

EvelynTSMG: I am of the solid belief that short single-line documentation comments SHOULD be folded onto one line along with the tags:
```cs
/// <summary>A numeric/spinbox input.</summary>
```

EvelynTSMG (changes requested):

src/core/settings.cs
fkelava: The contract of the method is that the out-var can be `null` depending on `T`. That it isn't in this case (``T = bool``, ``default(T) = false``) does not make it a generally sound way of calling it. I'd like not to cause any confusion.

src/core/settings.cs
fkelava: ID as a noun (abbreviation of 'identifier') is properly uppercase, lowercase 'id' is 'the deepest part of the unconscious mind' or whatever the exact definition of the psychological term is. I think it should stand.

EvelynTSMG: Hm, alrighty

EvelynTSMG (approved):

fkelava:
Despite the name, this PR did not 'finish' the mod settings API enablement. A followup will be filed in due course to fix some identified issues.

EvelynTSMG:
The name has been edited to better reflect the PR's scope.

PR #180, opened 2026-06-21
Make the Runtime appear like a SDK-based mod, and some other things

### LINKED ISSUE #179: The runtime should 'borrow' manifest autogeneration from the SDK
To appear as though it were any other mod, and to provide us with a concrete version identifier in the mod list and otherwise.

fkelava:
The runtime is special; it's an in-tree mod. Consequently, it cannot use the SDK.

That doesn't mean we don't want it to _look_ and _behave_ like an SDK-based mod; it's just that we have to manually keep the SDK and the runtime project in sync. But that's a minor enough chore.

Also included; a rewritten README and a shortened DLL name to run a smaller risk of running afoul of `MAX_PATH` in `state`.


fkelava:
Closes #179.

src/runtime/README.md
```diff
@@ -1,8 +1,22 @@
...
+- `platform.cs` - provides binding to any required native API (in our case, Win32+D3D11)
+- `imgui.cs` - provides for [Dear ImGui](https://github.com/ocornut/imgui) overlays over the game
+- `peloader.cs` - interfaces with the Phyre asset system to load game assets
+- `resloader.cs` - loads resources such as textures from disk, or through the game
+- `save_impl.cs` - a full reimplementation of the game's save system
+- `save_mgr.cs` - retrofits 'save set' capability over the game, and extends the save game limit
+- `save_ui.cs` - provides a ImGui user interface for the modified save system
+- `localstate.cs` - provides to each module a chance to persist data on save/load
+- `modconfig.cs` - provides a ImGui user interface for the mod configuration system
+- `lifecycle.cs` - fires callbacks to make modules aware of various occurrences in the game(s)
```
EvelynTSMG: More generally speaking, `events/*.cs`, not just `lifecycle.cs`.
Also, "callbacks" is the wrong word, as it implies you call a function that then uses the callback to, you know, call back to you? That is not the case for events.

EvelynTSMG (changes requested):
`fhr` is fine, though I don't think it's necessary, nor will it be. It saves 6|12 characters, which seems rather pointless.

src/runtime/README.md
```diff
@@ -19,4 +19,4 @@ The Runtime is broadly subdivided into:
 - `save_ui.cs` - provides a ImGui user interface for the modified save system
 - `localstate.cs` - provides to each module a chance to persist data on save/load
 - `modconfig.cs` - provides a ImGui user interface for the mod configuration system
-- `lifecycle.cs` - fires callbacks to make modules aware of various occurrences in the game(s)
+- `events/*.cs`, `lifecycle.cs` - fires events to make modules aware of various occurrences in the game(s)
```
EvelynTSMG, line 22: Are you going to mention every individual file in that folder here, when more are added? That's going to be quite a few, you know.

fkelava: Probably going to abridge it at some point. That said, we'll get there when we get there...

EvelynTSMG (approved):

PR #187, opened 2026-06-26
Use a custom-made managed ImGui wrapper

fkelava:
This is a draft and is not yet in a functional state.

One thing that [Hexa.NET.ImGui](https://github.com/HexaEngine/Hexa.NET.ImGui/) doesn't do, despite its general ease-of-use and robustness, is ship with symbols or debug binaries. This is not _normally_ a problem, as one could just generate a debug `cimgui` themselves for the matching version... in theory.

In practice, however, I have not succeeded in doing so despite my best efforts. It thus stands as the only thing which we cannot meaningfully debug in a complete Fahrenheit setup. Therefore, it may be time to go back to square one, where we came from- and compile our own ImGui. This time, though, we can do better than last time:
- We'll use [Dear Bindings](https://github.com/dearimgui/dear_bindings/) instead of [cimgui](https://github.com/cimgui/cimgui) to generate the "C API"
- We'll use our existing [ClangSharp](https://github.com/dotnet/ClangSharp) support to generate a managed wrapper over that.
- We'll make it so that one does not need to pull the ImGui submodules to _build_ Fahrenheit; building still works in one click using the version last committed to, and only when _updating_ does one have to run the new scripts


fkelava:
Aborted until further notice. The benefit is not worth the cost at this moment.

Issue #191, opened 2026-07-07
Localization and settings IDs should have the module type name implicitly prepended


fkelava:
The primary purpose of this is to prevent ID conflicts between settings. It is trivially possible to match category and setting names between modules by accident; it is even _reasonably likely_, on account of most mods having a common set of baseline options.

Another purpose is to standardize override addressing. To override the localization or probe the setting of another module, it would be useful (both to implementor and later reader) to know what an override is targeting immediately from the affected ID.

Finally, addressing by module type name is commonly used in Fahrenheit on-disk structures, and this merely aligns with that practice.

Consequently, localization and settings IDs (at least) will have the full module type name implicitly prepended, ex. ``Fahrenheit.Runtime.FhSaveUiModule.*``. Note that this alone does not suffice to _truly_ disambiguate a module; for that, we would also need the assembly name, since namespaces are free to collide in different assemblies. However, for present purposes, even ignoring that would provide a substantial gain.

fkelava:
Partially handled by #196. Settings still need this done.

fkelava:
Settings part deferred to next release while we hash out how to actually _get_ the caller's identity for this purpose.

PR #196, opened 2026-07-08
Rework internationalization support

### LINKED ISSUE #190: Localization should be split per-module, then per-language
Currently, mods can ship one file per language for all their localization. This is at odds with Fahrenheit's design, which encourages splitting up mods into discrete _modules_. Large mods would thus have unwieldy and large localization files despite a neater internal organization.

The two should be brought into alignment, localization taking the form:
```
/mods/{MOD_ID}/lang/{MODULE_TYPE_NAME}/{LANG_ID}.json
```
This also allows a natural expression of localization overrides, e.g.:
```
/lang/MyMod.MyModule/en_US.json                    -> own module
/lang/Fahrenheit.Runtime.FhSaveUiModule/en_US.json -> overrides Runtime
```

The SDK and in-tree build must be updated to match.

fkelava:
Fixes #190, #191.

Localization is now split per-module then per-language. Localization IDs get the declaring module's type implicitly prepended. To refer to another mod's localization string, it can be fully qualified; _or_ a `caller` parameter can be passed into `localize` to find the string with the given ID in the caller's "namespace".

See attached issues for further context.


src/core/i18n.cs
```diff
@@ -13,38 +13,84 @@ namespace Fahrenheit;
...
+            load_for_module(mod, module_lang_dir);
+        }
+    }
 
-                FhInternal.Log.Info($"Parsing locale {lang_id} for mod {mod_name}.");
+    /// <summary>
+    ///     Loads localization data for a given <paramref name="mod"/>'s module.
+    /// </summary>
+    internal void load_for_module(FhModContext mod, DirectoryInfo module_dir) {
+        foreach (FileInfo lang_file in module_dir.EnumerateFiles("*.json", SearchOption.TopDirectoryOnly)) {
```
EvelynTSMG, line 55: This would *work*, but it *should* be `*.lang.json`

src/core/i18n.cs
```diff
@@ -13,38 +13,84 @@ namespace Fahrenheit;
...
+             * The try-block is simply to provide a description of what locale faulted.
+             */
+            LocaleData user_locale;
+            try {
+                using FileStream lang_file_stream = lang_file.Open(FileMode.Open, FileAccess.Read);
+
+                user_locale = JsonSerializer.Deserialize<LocaleData>(lang_file_stream, FhUtil.InternalJsonOpts)
+                    ?? throw new Exception("Invalid or uninterpretable language file.");
+            }
+            catch {
```
EvelynTSMG, line 77: Remind me why we don't use the usual
```cs
if (...) {
    ...
} else {
    ...
}
```
logic in try-catch statements like so?
```cs
try {
    ...
} catch (...) {
    ...
}
```

src/core/i18n.cs
```diff
@@ -53,741 +99,13 @@ internal void initialize() {
     ///     for the locale with ID <paramref name="lang_id"/>,
     ///     falling back to <paramref name="id"/> if unavailable.
     /// </summary>
-    public string localize(string id, string lang_id = "en-US") {
-        return _localization_map.TryGetValue(lang_id, out LocaleData? locale) && locale.TryGetValue(id, out string? localized_string)
+    public string localize(string id, FhModule? caller = default, string lang_id = "en-US") {
+        string composite_id = caller == default
```
EvelynTSMG: The default for nullable values is `null`, and some IDEs (Rider) will suggest using `null` instead. I suggest that as well, since it makes the `caller == default` part easier to understand (as `caller == null` or `caller is null`).

src/runtime/Fahrenheit.Runtime.csproj
```diff
@@ -49,8 +49,7 @@
 
     <!-- ASSETS: LOCALIZATION -->
     <ItemGroup>
-        <None Remove="lang/**" />
-        <None Include="lang/*.json" CopyToOutputDirectory="Always" />
+        <None Include="lang/**/*.json" CopyToOutputDirectory="Always" />
```
EvelynTSMG, line 52: Should this not also use `*.lang.json`? I suppose it is inconsequential, but still. I think it would be easier for new mod devs to realize they misnamed the file by seeing VS didn't copy it than by the localization not working in-game.

src/sdk/Sdk/Sdk.props
```diff
@@ -24,8 +24,7 @@
 
     <!-- ASSETS: LOCALIZATION -->
     <ItemGroup>
-        <None Remove="lang/**" />
-        <None Include="lang/*.json" CopyToOutputDirectory="Always" />
+        <None Include="lang/**/*.json" CopyToOutputDirectory="Always" />
```
EvelynTSMG, line 27: Should this not also use `*.lang.json`? I suppose it is inconsequential, but still. I think it would be easier for new mod devs to realize they misnamed the file by seeing VS didn't copy it than by the localization not working in-game.

EvelynTSMG (changes requested):

src/core/i18n.cs
fkelava: Because I don't write `if` clauses that way to begin with, so I don't write try-catches that way either.

src/runtime/Fahrenheit.Runtime.csproj
fkelava: I didn't actually make it require the `*.lang.json` extension yet. This may change, but at least for now it's not strictly necessary.

EvelynTSMG (approved):

PR #199, opened 2026-07-08
Miscellaneous cleanup for `alpha11`

fkelava:


src/core/handles.cs
```diff
@@ -188,10 +188,10 @@ internal sealed class FhMethodContext {
...
 internal sealed class FhMethodTable {
 
-    private readonly Dictionary<nint,     Delegate>        _fnptrs     = []; // Any function -> cached delegate
-    private readonly Dictionary<nint,     FhMethodContext> _methods    = []; // Original     -> all hooks (for debug/keep-alive)
-    private readonly Dictionary<nint,     nint>            _chain_next = []; // Original     -> next chain insertion address
-    private readonly Dictionary<Delegate, nint>            _chain      = []; // Hook         -> next function in chain
+    private readonly static Dictionary<nint,     Delegate>        _s_fnptrs  = []; // Any function -> Cached delegate
+    private readonly static Dictionary<nint,     FhMethodContext> _s_methods = []; // Original     -> All hooks (for keep-alive)
+    private readonly static Dictionary<nint,     nint>            _s_insert  = []; // Original     -> Insertion address for next hook
+    private readonly static Dictionary<Delegate, nint>            _s_chain   = []; // Hook         -> Next function in chain
```
EvelynTSMG: Are we doing `s_` prefixes now?

fkelava: In the infrequent case of a static field appearing in a non-static type, I figured the distinction is worth showing at a glance. I don't overly care.

EvelynTSMG: We don't do them elsewhere, and since these are private it doesn't really matter, so I think we should avoid it for consistency. As a bonus it would make the GitHub diff cleaner lol.

EvelynTSMG: We don't do `s_` prefixes anywhere else, and since these are private it doesn't really matter, so I think we should avoid it for consistency. As a bonus it would make the GitHub diff cleaner lol.

EvelynTSMG (changes requested):

EvelynTSMG (dismissed):

EvelynTSMG (approved):

Issue #140, opened 2026-05-07
Modal confirmation UI/permission requests for mods


fkelava:
Some actions, such as save set manipulation, should not be available to mods without soliciting the player's input. The player must be able to pre-empt the mod, and silence it if it is being obstinate against their wishes.

This requires modal confirmation dialogs to be presented to the user. We should decide on their form, specific details such as the ability to 'mute' mods' repeated requests (whether manually or automatically), and decide what Core API to expose for this feature.

PR #202, opened 2026-07-14
Lift the save manager and local state machinery into Core

### LINKED ISSUE #201: Sets and state are purely runtime concepts, but should be Core API instead
The concept of save sets and local state in Fahrenheit are entwined. Both are implemented in the runtime, and do not appear as part of Core API. See:
https://github.com/fahrenheit-crew/fahrenheit/blob/e11c74db502c513c4956163ca9514156523b4408/src/runtime/localstate.cs#L30-L56

Strictly speaking, _nothing at all_ demands these concepts be in the runtime. When the save system rewrite was broken down, this was merely done for implementation expediency. If we want #111 and similar to exist at all, then they must be lifted out of the runtime.

fkelava:
Fixes #201.

Note that this still does not give sets proper Core 'API'- that is still blocked on modal permission requests (see #140), but at least it clarifies where these pieces properly belong.

### DIFF
--- a/src/core/save_mgr.cs
+++ b/src/core/save_mgr.cs
@@ -8,7 +8,6 @@ namespace Fahrenheit;
 /// <summary>
 ///     Allows multiple sets of saves to exist.
 /// </summary>
-[FhLoad(FhGameId.FFX | FhGameId.FFX2 | FhGameId.FFX2LM)]
 internal sealed class FhSaves {
 
     /* [fkelava 07/11/25 15:01]
--- a/src/core/save_mgr.cs
+++ b/src/core/save_mgr.cs
@@ -3,15 +3,13 @@
 // This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
 // It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
 
-using System.Diagnostics;
-
-namespace Fahrenheit.Runtime;
+namespace Fahrenheit;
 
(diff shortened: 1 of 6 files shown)


src/core/save_mgr.cs
```diff
@@ -3,15 +3,13 @@
...
 
-using System.Diagnostics;
-
-namespace Fahrenheit.Runtime;
+namespace Fahrenheit;
 
 /// <summary>
 ///     Allows multiple sets of saves to exist.
 /// </summary>
 [FhLoad(FhGameId.FFX | FhGameId.FFX2 | FhGameId.FFX2LM)]
```
EvelynTSMG: You kept the `FhLoad` attribute?

src/core/state.cs
```diff
@@ -99,12 +88,13 @@ internal void state_save_slot(int slot) {
...
+                FhInternal.Log.Debug($"Writing metadata for module {module_type}.");
                 state_meta_fs.Write(JsonSerializer.SerializeToUtf8Bytes(state_meta, FhUtil.InternalJsonOpts));
 
-                _logger.Debug($"{module_type} -> {state_fn}.");
+                FhInternal.Log.Debug($"{module_type} -> {state_fn}.");
                 module_context.Module.save_local_state(state_fs);
             }
         }
     }
+
```
EvelynTSMG: Why was this empty line added?

EvelynTSMG (changes requested):

EvelynTSMG (approved):

PR #203, opened 2026-07-14
Lift ImGui vtable delegates into Core

fkelava:
There is no particular reason why these should be private to the ImGui module. The fact that they refer to vtable functions does not preclude their 'handleization' in any way.


EvelynTSMG (approved):
I have vague stylistic concerns but I've had them for a while and we shouldn't bother addressing them until I start writing the style guide.

PR #204, opened 2026-07-14
More miscellaneous cleanup for `alpha11`

fkelava:
STEP is an internal development tool that is of no use to consumers of Fahrenheit packages. We should not be shipping it.

Additionally, restrict and clarify an overbroad `unsafe` in texture API.

### DIFF
--- a/src/core/resload.cs
+++ b/src/core/resload.cs
@@ -3,8 +3,6 @@
 // This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
 // It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
 
-using Hexa.NET.ImGui;
-
 namespace Fahrenheit;
 
 /// <summary>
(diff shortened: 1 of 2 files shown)


src/core/resload.cs
```diff
@@ -109,13 +107,18 @@ internal void load(ImTextureRef texture_ref, FhTextureMetadata texture_metadata)
...
+    internal void unload() {
         /* [fkelava 01/05/26 18:51]
-         * Testing the return value is meaningless because it is not guaranteed to be precise.
-         *
+         * Testing the return value of Release is meaningless because it is not guaranteed to be precise.
          * If you believe you're leaking textures, turn on the D3D debug layer instead.
+         *
+         * SAFETY: cast of opaque 'ImTextureID' to true underlying type 'ID3D11ShaderResourceView' is well defined
+         * see: https://github.com/ocornut/imgui/blob/da137cbbb066e57f4b44f3f331c36e1e30e1cbe2/backends/imgui_impl_dx11.cpp#L5
+         * > {...} Use 'ID3D11ShaderResourceView*' as texture identifier. {...}
```
EvelynTSMG: While I agree with adding this, these should be separate comments as they differ in topic.

EvelynTSMG (changes requested):

EvelynTSMG (approved):

PR #205, opened 2026-07-16
Rework STEP and regenerate delegates

fkelava:
Per some Discord discussions.

- Prefix delegates with `d_`
- No longer prefix hooks with `h_`
- Provide STEP data as part of source to allow STEP regen
- Provide for a "no-emit" file to avoid emitting manualized calls


EvelynTSMG (approved):
Looks good to me!

PR #208, opened 2026-07-16
Yet more miscellaneous cleanup for `alpha11`

fkelava:
- Replace `fixed` fields with equivalent inline array declarations
- Shove all inline-array types in `src/core/inlarrays.cs`
- Document this and that in the save PAL
- Torch an unused file entirely
- Test layout is meant to correspond to `src`; let's make that happen again for `encoding`


src/core/inlarrays.cs
```diff
@@ -0,0 +1,76 @@
...
+
+/* [fkelava 23/9/25 18:56]
+ * For sizes up to 16, the standard library provides inline array types.
+ * We use those where possible. For larger sizes, we define our own here.
+ */
+
+[InlineArray(17)]
+public struct InlineArray17<T> {
+    private T _t;
+}
```
EvelynTSMG, line 18: As per discord, please leave a
```
//TODO: Change uses of this type to new custom types, one for each purpose this type has
```
comment before every generic type you made to remove `fixed`.

EvelynTSMG (changes requested):

EvelynTSMG (approved):
Fine

PR #212, opened 2026-07-18
Convert existing uses of `FhEnumExt` to extension properties

### LINKED ISSUE #130: Provide extension properties to `[Flags]` enums
This should follow the style:
```cs
public static partial class FhEnumExt {
    extension(EnumName flags) {
        public bool flag1 { readonly get { return flags.get_bit((int)EnumName.FLAG_1); } set { flags.set_bit((int)EnumName.FLAG_1, value); } }
        public bool flag2 { readonly get { return flags.get_bit((int)EnumName.FLAG_2); } set { flags.set_bit((int)EnumName.FLAG_2, value); } }
        public bool flag3 { readonly get { return flags.get_bit((int)EnumName.FLAG_3); } set { flags.set_bit((int)EnumName.FLAG_3, value); } }
        public bool flag4 { readonly get { return flags.get_bit((int)EnumName.FLAG_4); } set { flags.set_bit((int)EnumName.FLAG_4, value); } }
    }
}
```

\
This issue includes both converting old extension methods in `FhEnumExt` (as visible in [`status.cs`](https://github.com/fahrenheit-crew/fahrenheit/blob/95a0e57d49405678e3f458aec347ca0e3194c53b/core/fh/FFX/status.cs)), as well as adding new ones for the relevant flags: for example, in [`command.cs`](https://github.com/fahrenheit-crew/fahrenheit/blob/95a0e57d49405678e3f458aec347ca0e3194c53b/core/fh/FFX/command.cs).

fkelava:
Handles the 'convert existing' part of #130.

### DIFF
--- a/src/core/ffx/aability.cs
+++ b/src/core/ffx/aability.cs
@@ -55,22 +55,77 @@ public enum StatIncreaseFlags : ushort {
 
 public static partial class FhEnumExt {
     extension(StatIncreaseFlags flags) {
-        public bool strength      => flags.HasFlag(StatIncreaseFlags.STRENGTH);
-        public bool defense       => flags.HasFlag(StatIncreaseFlags.DEFENSE);
-        public bool magic         => flags.HasFlag(StatIncreaseFlags.MAGIC);
-        public bool magic_defense => flags.HasFlag(StatIncreaseFlags.MAGIC_DEFENSE);
-        public bool agility       => flags.HasFlag(StatIncreaseFlags.AGILITY);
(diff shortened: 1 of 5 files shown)


EvelynTSMG (changes requested):
As per the original issue, please add `set` accessors for setting the flags as well.
This is annoying, but the issue outlines how it should look and you should be able to use find-and-replace for it.
Thank you for your hard work!

EvelynTSMG (approved):
Good work!

PR #214, opened 2026-07-20
ModMgr: Switch program stub to OGL3+SDL3

fkelava:
All of our tools with graphical interfaces rely on the concept of a 'program stub' and an ImGui-based UI. The idea is simple; edit one method, and all the other parts of _being a functional program_ are taken care of by the stub, which is shared among all tools.

This 'program stub' was Win32+D3D11, just like the game itself. This was so we could share DLLs we already bring into the game for our tools. This permits them to run under emulation on Linux systems, but that's not an ideal user experience. While we can't do anything about the game, tools can move to a cross-platform program stub.

So let's try switching that out for an OpenGL3+SDL3 program stub.

It mostly works, but there are open issues that still make this a draft:
- How do we package the things that are now only used for these tool program stubs?
- In release, we have a spurious fail-fast due to a stack buffer overrun. This is somewhere in the bowels of Hexa.NET's packages and thus virtually undebuggable by normal means.


fkelava:
The underlying issue was debugged to be a temporary break in Hexa.NET's x86 OGL/SDL support.

We are torching hard-forced bitness anyway (see #215), so this is ready for review.

src/modmgr/main.cs
```diff
@@ -10,330 +10,217 @@ namespace Fahrenheit.Tools.ModManager;
...
+        SDLWindowFlags window_flags =
+            SDLWindowFlags.Opengl
+          | SDLWindowFlags.Resizable
+          | SDLWindowFlags.Hidden
+          | SDLWindowFlags.HighPixelDensity;
+
+        _sdl_window = SDL.CreateWindow(
+            "Fahrenheit Mod Manager",
+            int.CreateChecked(1280 * main_display_scale),
+            int.CreateChecked(800  * main_display_scale),
```
EvelynTSMG, line 72: 1280x800 is somewhat interesting, but I don't have a problem with it. Any anxieties I *do* have I could only dispel by researching for longer than I care to, unfortunately.

src/modmgr/main.cs
```diff
@@ -10,330 +10,217 @@ namespace Fahrenheit.Tools.ModManager;
...
 
-        if (_p_swap_chain != null) { _p_swap_chain->Release(); _p_swap_chain = null; }
-        if (_p_device_ctx != null) { _p_device_ctx->Release(); _p_device_ctx = null; }
-        if (_p_device     != null) { _p_device    ->Release(); _p_device     = null; }
-    }
+        SDL.GLMakeCurrent    (_sdl_window, _sdl_context);
+        SDL.GLSetSwapInterval(1); // Enable VSync
+        SDL.SetWindowPosition(_sdl_window,
+            int.CreateChecked( SDL.SDL_WINDOWPOS_CENTERED_MASK | 0 ),
+            int.CreateChecked( SDL.SDL_WINDOWPOS_CENTERED_MASK | 0 ));
```
EvelynTSMG: Surely `| 0` does nothing here??
Moreover, `SDL.SDL_WINDOWPOS_CENTERED_MASK` is defined as `0x2FFF0000u`, which is an *unsigned* integer, so using `int.CreateChecked` seems rather weird.

src/modmgr/main.cs
```diff
@@ -10,330 +10,217 @@ namespace Fahrenheit.Tools.ModManager;
...
         }
 
-        // ImGui setup
-        ImGuiContextPtr ctx   = ImGui.CreateContext();
-        ImGuiIOPtr      io    = ImGui.GetIO();
-        ImGuiStylePtr   style = ImGui.GetStyle();
+        // Cleanup
+        ImGuiImplOpenGL3.Shutdown();
+        ImGuiImplSDL3   .Shutdown();
+        ImGui           .DestroyContext();
```
EvelynTSMG: Can we add a blank line here so the latter line can be `ImGui.DestroyContext()`? 🥺

EvelynTSMG (changes requested):

src/modmgr/main.cs
EvelynTSMG: To be clear, this does not need changing, I just added it to the review instead of commenting by accident.

src/modmgr/main.cs
fkelava: The function only accepts an `int` while the constant was defined as `uint`, one of the many unfortunate details about this library combo. It's that or a cast, I _suppose_ I can cast.

EvelynTSMG (approved):
Looks good to me! Thank you for all the work you've been putting in :3

Issue #215, opened 2026-07-20
Fahrenheit and its tools should be bitness-agnostic, and packaged separately for supported platforms


fkelava:
Fahrenheit, and its tools, today ship with a forced `RuntimeIdentifier` of `win-x86`. This is a natural consequence of the fact the only supported version of the game ships as a Windows 32-bit binary. Since the game runs on Unices of all kinds through emulation, Fahrenheit and its tools do too.

This has worked so far, but it's not without issues. Put most simply, emulation is bothersome, and results in a much worse debugging experience than running natively. Furthermore, Fahrenheit and its tools are mostly platform- and bitness-agnostic; it is only when injecting into the game, for a certain subset of types, that the 32-bit requirement applies.

More pertinently, it is simply better to offer cross-platform tools. But since tools link against Fahrenheit, they cannot move to AnyCPU, 64-bit builds until Fahrenheit itself no longer has a mandated RID.

So we should lift that restriction. This entails:
- Manually checking every pointer-type and `nint` usage for safety in the face of AnyCPU compilation and usage
- Manually retooling the build system and CI to produce 'neat' packages for supported platform triples:
  - `win-x86`
  - `win-x64`
  - `linux-x64`
- Manually amending loader code to function in the face of different packaging of native assets.

See also generally #71.

fkelava:
Note that in any package/event, `fhr`, `fhstage0` and `fhstage1` will continue shipping only as `win-x86` binaries and require emulation. This will not be lifted until the game itself is updated or we support more than one platform.

PR #216, opened 2026-07-20
EEdit: Switch program stub to SDL3+OGL3

fkelava:
Contributes to #215.


EvelynTSMG (approved):
I skimmed it. Seems good!

fkelava:
It's materially identical to #214 (`main.cs` is copied verbatim sans the namespace), so skimming suffices here.

PR #217, opened 2026-07-20
Support `win-x86`, `win-x64`, and `linux-x64` RIDs seamlessly

### LINKED ISSUE #215: Fahrenheit and its tools should be bitness-agnostic, and packaged separately for supported platforms

fkelava:
Closes #215.

Fahrenheit and its tools should now no longer be locked to `win-x86` mode, allowing native tool builds on Linux and 64-bit tool builds on Windows.

### DIFF
--- a/.github/workflows/msbuild.yml
+++ b/.github/workflows/msbuild.yml
@@ -26,11 +26,7 @@ env:
 
 jobs:
   build:
-    strategy:
-      matrix:
-        architecture: [ x86 ]
-        os: [ windows ]
-    runs-on: ${{ matrix.os }}-2025
+    runs-on: windows-2025
 
     steps:
     - name: Print runner environment
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -52,7 +52,6 @@
         </ItemGroup>
     </Target>
 
-
     <!-- fkelava 09/06/26 17:36
     A placeholder target to force publishing to occur after a build.
     Publishing places build output in its proper place.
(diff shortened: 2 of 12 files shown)


fkelava:
On Linux systems, `dotnet {TOOL}.dll` will launch a native version of the given binary. Fahrenheit packages are thus now "native anywhere" with the exception of `fhstage0` and `fhstage1`, which still run under emulation, just like the game itself.

EvelynTSMG (approved):

PR #218, opened 2026-07-20
Raise `vcpkg` baseline, extend boot prints

fkelava:
We haven't updated our `vcpkg` baseline since we first switched to it. This unlocks the use of newer `detours`, `nethost` and `minhook` binaries.


EvelynTSMG (approved):

EvelynTSMG:
Out of curiosity, do newer `minhook` binaries provide any new features or API surface we could benefit from?

fkelava:
Not that I know of. This is primarily because we were using a rather old `nethost`, which is not great when we need access to recently introduced properties like `APP_CONTEXT_BASE_DIRECTORY` in #217.

PR #220, opened 2026-07-21
Lift debug delegates into Core

fkelava:
See https://github.com/fahrenheit-crew/fh-mods-debug/pull/5.

### DIFF
--- a/src/core/call.cs
+++ b/src/core/call.cs
@@ -234,7 +234,7 @@ public static unsafe partial class FhCall {
     [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
     public delegate float d_CT_RetFloat(AtelBasicWorker* work, int* storage, AtelStack* stack);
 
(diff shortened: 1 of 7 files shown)


src/core/call.cs
```diff
@@ -234,6 +234,43 @@ internal static FhMethodHandle<d_SaveDataSaveLoadSucceed> SaveDataSaveLoadSuccee
     [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
     public delegate float d_CT_RetFloat(AtelBasicWorker* work, int* storage, AtelStack* stack);
 
+    // Debug mod - `printf` style methods
```
EvelynTSMG:
```suggestion
    // `printf` and similar methods for use by the debug mod
```

EvelynTSMG (changes requested):

EvelynTSMG (approved):

PR #221, opened 2026-07-21
Do not count the auto-save for save slot limit

### LINKED ISSUE #87: Do not include autosave in the used and maximum slot counts in the save UI
This is confusing when the player reaches the first save sphere and sees "1/500". I recommend simply displaying "0/499".

EvelynTSMG:
... by increasing the maximum amount of save slots to 501 and then making the UI pretend that isn't the case.

Resolves #87.

### DIFF
--- a/src/runtime/save_ui.cs
+++ b/src/runtime/save_ui.cs
@@ -83,7 +83,8 @@ public sealed class FhSaveUiModule : FhModule {
 
         ReadOnlySpan<FhSaveDisplayData> display_data = FhInternal.Saves.get_display_data();
 
-        for (; slot < FhInternal.Saves.get_slots_total(); slot++) {
+        // get_slots_total() does not count the auto-save by default, so we have to add 1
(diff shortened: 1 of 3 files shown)


src/runtime/save_ui.cs
```diff
@@ -83,7 +83,8 @@ private void ui_mainwindow() {
 
         ReadOnlySpan<FhSaveDisplayData> display_data = FhInternal.Saves.get_display_data();
 
-        for (; slot < FhInternal.Saves.get_slots_total(); slot++) {
+        // get_slots_total() does not count the auto-save by default, so we have to add 1
+        for (; slot < FhInternal.Saves.get_slots_total() + 1; slot++) {
```
EvelynTSMG: Alternatively, we could add a `get_slots_total_no_autosave()` method, revert this change, and use the new method in the other places `get_slots_total()` is used.

fkelava: I think we should go with this.

fkelava (approved):
Looks good, thank you.

PR #224, opened 2026-07-22
SaveMgr: Unrestrict set size, simplify contract

fkelava:
In #221 we went to fix a bug with the save manager including the autosave in the count. Due to shoddy error-checking on my behalf, that didn't quite work. While fixing it, it occurred to me that the whole thing could not only be _definitively_ fixed, but also vastly simplified by simply _not having a limit on set size in the first place_.

Note that this PR does not feature the kind of optimization work that allows extremely large save sets (1,000+) to be indexed rapidly. This is purely so we can simplify the manager and more trivially reason around it, and we can devote resources to making the UI and loading perform and scale better later.

### DIFF
--- a/src/core/save_mgr.cs
+++ b/src/core/save_mgr.cs
@@ -16,28 +16,21 @@ internal sealed class FhSaves {
      * At runtime, you may swap between all sets for the hash.
      */
 
-    private          int                            _sm_lock;
-    private readonly HashSet<string>                _sm_sets;
-    private readonly int                            _sm_set_size;
-    private          string                         _sm_active_set;
-    private readonly int[]                          _sm_active_set_slots;
-    private readonly int[]                          _sm_active_set_saves;
-    private          int                            _sm_active_set_count;
-    private readonly string                         _sm_path_base;
-    private readonly string                         _sm_path_default_set;
-    private readonly FhSaveDisplayData[]            _sm_display_data;
-    private readonly Dictionary<string, FileInfo[]> _sm_file_attributes;
+    private          int                     _sm_lock;
+    private readonly HashSet<string>         _sm_sets;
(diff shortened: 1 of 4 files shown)


fkelava:
Worry about the performance was a bit premature. As long as you're not running a debug build, sets exceeding 1,000 saves still index and display relatively rapidly. The UX becomes questionable, but at least _performance_ remains acceptable, with potential for considerable improvement left in the future if needed.

src/core/save_mgr.cs
```diff
@@ -233,56 +195,20 @@ internal string get_save_path_for_slot(int slot) {
     ///     Get the number of used up slots in the current set.
```
EvelynTSMG: The term "used up" should now be "used", as there is no longer a finite amount of slots.

src/runtime/save_ui.cs
```diff
@@ -219,7 +217,7 @@ private void ui_save_info_generic(FhSaveDisplayData data, int slot) {
 
         bool is_autosave = slot == 0 && _sem!.get_system_state() is FhSaveExtensionSystemState.LOAD;
```
EvelynTSMG: Since we show the autosave when synchronizing in an Al Bhed sphere now, I believe this needs to be
```suggestion
        bool is_autosave = slot == 0 && _sem!.get_system_state() is not FhSaveExtensionSystemState.SAVE;
```

EvelynTSMG (changes requested):
In addition to the requested changes, I believe we should consider making the `List<FhSaveDisplayData> _sm_display_data` field of the save manager an implementation detail in `save_ui.cs` (along with any menu indices), if possible.

EvelynTSMG (approved):

PR #231, opened 2026-07-27
Mod Manager Initial Design

foresterfx:
Supports:
- Read/Write operations on `loadorder` through the UI interactions
- Launching FFX, FFX-2, FFX-2LM via menu options (debug as well)
- Warning information given to user if binaries not found within game folder
- Importing/Exporting of zipped directory containing mods


EvelynTSMG:
... Could you at least tell the "AI" to be straightforward and to-the-point instead of talking as if it's trying to sell me a car?

My brain wants to explode whenever I try to extract *any* information out of this stars-forsaken text, so if it's not easy to read, I'm not doing so. Especially when you didn't care to write it in the first place.

src/modmgr/main.cs
```diff
@@ -60,23 +68,55 @@ private static void Main(string[] args) {
...
+        int window_width  = int.CreateChecked(Math.Clamp(usable_bounds.W * 0.65F, 1024F * main_display_scale, 2200F * main_display_scale));
+        int window_height = int.CreateChecked(Math.Clamp(usable_bounds.H * 0.70F, 700F  * main_display_scale, 1400F * main_display_scale));
+
         SDLWindowFlags window_flags =
             SDLWindowFlags.Opengl
           | SDLWindowFlags.Resizable
           | SDLWindowFlags.Hidden
-          | SDLWindowFlags.HighPixelDensity;
+          | SDLWindowFlags.HighPixelDensity
+          | SDLWindowFlags.Borderless;
```
fkelava: I notice that the choice to go borderless adds quite a bit of code to the initial proposal. For now, let's keep it simple and _not_ do that- remove `chrome.cs` as well and any comments and concessions done specifically for use in borderless mode.

Ex. L112-L119, L196-205, L263-274 in this file. Take those (and any other similar occurrences across the codebase) and save them for a follow-up PR.

fkelava (changes requested):

src/modmgr/packs.cs
```diff
@@ -0,0 +1,377 @@
...
+internal sealed record FhWritableManifest(
+    string Id,
+    string Name,
+    string Desc,
+    string Authors,
+    string Version,
+    string Link,
+    string[] Dependencies,
+    string[] LoadAfter,
+    string Flags);
```
fkelava: In hindsight, it was a mistake of mine that the mod manager does not depend on Fahrenheit.

#233 has been filed to address this. When it merges, rebase the PR and use the actual Fahrenheit type.

Do the same where-ever appropriate.

src/modmgr/theme.cs
```diff
@@ -0,0 +1,328 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+/* Jobs:
+ * - FhThemeColor: a JSON-friendly RGBA color used to persist theme overrides
+ *   in fhmodmgr.json (a plain System.Numerics.Vector4 doesn't round-trip
+ *   through the settings store's JSON options).
```
fkelava: Your `JsonSerializerOptions` likely didn't `IncludeFields`.

src/modmgr/theme.cs
```diff
@@ -0,0 +1,328 @@
...
+        style.Colors[(int)ImGuiCol.TableRowBgAlt]     = new Vector4(COLOR_TEXT.X, COLOR_TEXT.Y, COLOR_TEXT.Z, 0.03F);
+
+        style.Colors[(int)ImGuiCol.ScrollbarBg]          = COLOR_BACKGROUND;
+        style.Colors[(int)ImGuiCol.ScrollbarGrab]        = new Vector4(0.28F, 0.27F, 0.34F, 1.00F);
+        style.Colors[(int)ImGuiCol.ScrollbarGrabHovered] = new Vector4(0.37F, 0.35F, 0.45F, 1.00F);
+        style.Colors[(int)ImGuiCol.ScrollbarGrabActive]  = COLOR_ACCENT;
+
+        style.Colors[(int)ImGuiCol.Text]         = COLOR_TEXT;
+        style.Colors[(int)ImGuiCol.TextDisabled] = COLOR_TEXT_MUTED;
+    }
```
fkelava: Once #233 merges, this should use the style defined in Fahrenheit's `FhImGuiHelper`. We may have to slightly refactor the helper to allow its use outside of the game process to do so. Try it- if push comes to shove we'll do it ourselves and you'll rebase against it.

src/modmgr/theme.cs
```diff
@@ -0,0 +1,328 @@
...
+        ImGui.PushStyleColor(ImGuiCol.ButtonHovered, hovered_color);
+        ImGui.PushStyleColor(ImGuiCol.ButtonActive, active_color);
+
+        bool pressed = ImGui.Button(label, size);
+
+        ImGui.PopStyleColor(3);
+
+        return pressed;
+    }
+}
```
fkelava: Depending on @EvelynTSMG's opinion, we may want to move this up somewhere in or around the `ImGuiHelper`. Reusable elements shared in- and out of game seem like something we'll eventually have to get to.

fkelava (changes requested):

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,175 @@
...
+ * - FhShell: open an arbitrary folder in the OS's file browser (Explorer on
+ *   Windows, xdg-open elsewhere).
+ */
+
+namespace Fahrenheit.Tools.ModManager;
+
+internal enum FhLaunchTarget {
+    FFX,
+    FFX2
+}
```
fkelava: Likely this should be replaced with `FhGameId` post #233. The mod manager has to be able to launch Last Mission too. You can do that by running ``FFX-2.exe`` with launch arg `FFX2_LASTMISSION`.

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,175 @@
...
+        fahrenheit_bin,
+        "fhstage0.exe");
+
+        string stage1_path = Path.Join(
+        fahrenheit_bin,
+        "fhstage1.dll");
+
+        string game_path = Path.Join(
+        game_directory,
+        game_executable);
```
fkelava: Indent subsequent lines by 4 spaces, per `.editorconfig`.

```cs
string game_path = Path.Join(
    game_directory,
    game_executable);
```

Fix this up anywhere else it may be required.

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,175 @@
...
+                missing_files.Add(required_path);
+            }
+        }
+
+        if (missing_files.Count > 0) {
+            return new(
+                false,
+                "The Fahrenheit installation is incomplete.\n\n"
+                + "Missing files:\n"
+                + string.Join('\n', missing_files));
```
fkelava: Never concatenate strings with `+`. Use a format string instead, ex.
```cs
return new(false, $"The Fahrenheit installation is incomplete.\n\nMissing files:\n{string.Join('\n', missing_files)}");
```
Optionally, where the string is too long, a multiline string literal would be nicer. (``"""``)

Same anywhere else it may be found.

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,175 @@
...
+                false,
+                "The configured game directory is invalid.\n\n"
+                + exception.Message);
+        }
+
+        string game_executable = target switch {
+            FhLaunchTarget.FFX  => "FFX.exe",
+            FhLaunchTarget.FFX2 => "FFX-2.exe",
+            _ => throw new ArgumentOutOfRangeException(nameof(target))
+        };
```
fkelava, line 32: In Fahrenheit one generally aligns a `switch` expression by `=>`, ex.
```cs
string game_executable = target switch {
    FhLaunchTarget.FFX  => "FFX.exe",
    FhLaunchTarget.FFX2 => "FFX-2.exe",
    _                   => throw new ArgumentOutOfRangeException(nameof(target))
};
```

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,175 @@
...
+    bool Success,
+    string Message
+);
+
+internal static class FhGameLauncher {
+
+    internal static FhLaunchResult launch(
+        FhLaunchTarget target,
+        string configured_game_directory,
+        bool debug = false) {
```
fkelava: Align multiline function declarations/signatures like ex. https://github.com/fahrenheit-crew/fahrenheit/blob/main/src/core/encoding.cs#L776-L781.

Best do it by hand. The rule is not encoded in `.editorconfig` or a linter, we do it by hand.

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,175 @@
...
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+/* Jobs:
+ * - FhGameLauncher: validate the Fahrenheit install (stage0/stage1/game exe
+ *   all present) and launch FFX/FFX-2 through fhstage0.exe, optionally in
+ *   debug mode.
+ * - FhShell: open an arbitrary folder in the OS's file browser (Explorer on
+ *   Windows, xdg-open elsewhere).
+ */
```
fkelava: Everywhere they are found, convert these `Jobs` blocks left by the LLM (?) into docstrings on the actual types they refer to, ex.
```cs
/// <summary>
///     Validates the Fahrenheit install and launches the games.
/// </summary>
internal static class FhGameLauncher {
```
In so doing, feel free to make them more terse. It doesn't matter _how_, but _what it generally does_. Format them as other docstrings in Fahrenheit are, i.e. indent by 4 spaces, use `<br/>` or `<para/>` to break down multiline docstrings to ensure legibility.

fkelava (changes requested):

src/modmgr/icon.cs
```diff
@@ -0,0 +1,75 @@
...
+            GLPixelFormat.Rgba,
+            GLPixelType.UnsignedByte,
+            image.Data);
+
+        gl.BindTexture(GLTextureTarget.Texture2D, 0);
+
+        Texture = new ImTextureRef(null, new ImTextureID((ulong)texture_name));
+        Size    = new Vector2(image.Width, image.Height);
+    }
+}
```
fkelava: In line with my request to not ship borderless support at first, remove this entire file.

Note how the project file for the mod manager already contains an `<ApplicationIcon>` attribute. The icon already showed on the system titlebar, at least on Windows.

src/modmgr/Fahrenheit.Tools.ModManager.csproj
```diff
@@ -39,6 +39,26 @@
...
+    </ItemGroup>
+
+    <!-- ICON ASSETS -->
+    <ItemGroup>
+        <Content Include="..\..\assets\fh_base_256.png">
+            <Link>resources\icons\fh_base_256.png</Link>
+            <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+            <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+        </Content>
     </ItemGroup>
```
fkelava: See the comment at `icon.cs`. This need not exist in that case.

fkelava (changes requested):

src/modmgr/launcher.cs
fkelava: Note also that I would then expect 'Play' to feature options for LM.

EvelynTSMG:
Please click the "Resolve Conversation" button on GitHub when fulfilling a change request. It makes it much easier to scroll through the PR later. I have taken the liberty of doing so myself for the ones you reacted with :+1: to.

src/modmgr/mods.cs
```diff
@@ -0,0 +1,394 @@
...
+        }
+
+        string fahrenheit_directory = Directory.GetParent(AppContext.BaseDirectory)!.FullName;
+        string mods_directory       = Path.Join(fahrenheit_directory, "mods");
+
+        if (!Directory.Exists(mods_directory) && Directory.Exists(Path.Join(game_directory, "fahrenheit", "mods"))) {
+            warnings.Add("The Fahrenheit mods directory was not found in the application directory. Checking the game directory instead.");
+            warnings.Add("If you installed Fahrenheit to a custom location, please move the bin directory to 'game directory/fahrenheit'");
+            mods_directory = Path.Join(game_directory, "fahrenheit", "mods");
+        }
```
fkelava, line 57: This is wrong. The mod manager is emitted to `bin` in a normal build (and runs from there), so `AppContext.BaseDirectory` will be `{GAME_DIR}\fahrenheit\bin`, and you'll incorrectly try to find mods under `{GAME_DIR}\fahrenheit\bin\mods`.

fkelava (changes requested):

src/modmgr/ui_efl_import_modal.cs
```diff
@@ -0,0 +1,121 @@
...
+        ImGui.RadioButton("Final Fantasy X-2##EflGame", ref _efl_import_game_index, 1);
+
+        ImGui.Spacing();
+
+        ImGui.Text("Mod ID");
+        ImGui.SetNextItemWidth(-1F);
+        ImGui.InputText("##EflImportModId", ref _efl_import_mod_id, MOD_ID_INPUT_LENGTH);
+        ImGui.Text("Mod name (optional, defaults to the ID)");
+        ImGui.SetNextItemWidth(-1F);
+        ImGui.InputText("##EflImportModName", ref _efl_import_mod_name, MOD_ID_INPUT_LENGTH);
```
fkelava, line 44: One should also be permitted to specify an author, version, and link. All three can be text inputs as well.

src/modmgr/packs.cs
```diff
@@ -0,0 +1,297 @@
...
+        }
+
+        try {
+            string efl_subfolder = game == FhGameId.FFX2 ? "x2" : "x";
+            string efl_directory = Path.Join(mod_directory, "efl", efl_subfolder);
+
+            Directory.CreateDirectory(efl_directory);
+
+            _copy_directory(source_folder, efl_directory);
+
```
fkelava, line 43: This is insufficiently constrained. If the user gives the wrong folder, import will proceed regardless and the resulting mod will have no effect. An EFL mod may only contain (at its root) one or more of the following subfolders:

- `ffx_data`
- `ffx_ps2`
- `ffx-2_data`
- `metamenu`
- `version_config`

Check case-insensitively.

fkelava (changes requested):

src/modmgr/ui_menu.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            ImGui.EndMenu();
+        }
+
+        if (ImGui.MenuItem("Settings")) {
+            _show_settings_dialog = true;
+        }
+
+        Vector2 menu_cluster_max = ImGui.GetItemRectMax();
+
```
fkelava, line 92: I take it this was overlooked when erasing some code. It's unused.

src/modmgr/launcher.cs
```diff
@@ -0,0 +1,113 @@
...
+
+        foreach (string required_path in new[] { stage0_path, stage1_path, game_path }) {
+            if (!File.Exists(required_path)) {
+                missing_files.Add(required_path);
+            }
+        }
+
+        if (missing_files.Count > 0) {
+            return new(false, $"Fahrenheit installation is incomplete.\n\nMissing files:\n{string.Join('\n', missing_files)}");
+        }
```
fkelava, line 48: This check is well-intentioned, but a bit pointless. If `fhstage1.dll` or the game binary were missing, then Stage0 would surface that error. If `fhstage0` were missing, the try-block beneath would catch that. Besides, quite a few more files are required for boot to properly succeed.

We can just do away with this and include a "health check" function in a follow-up PR.

src/modmgr/mods.cs
```diff
@@ -0,0 +1,394 @@
...
+    ///    If the mod directory or manifest is missing or malformed, the returned record will report error details.
+    /// </summary>
+    private static FhInstalledMod _read_mod(string mods_directory, string mod_id, int? load_order_index) {
+        string mod_directory = Path.Join(mods_directory, mod_id);
+        // guard against mod directory not existing after it was enumerated
+        bool directory_exists = Directory.Exists(mod_directory);
+
+        if (!directory_exists) {
+            return new(new FhManifest(mod_id, mod_id, "", "", "", "", [], [], FhManifestFlags.NONE), mod_directory, false, false, "The mod directory does not exist.", load_order_index);
+        }
```
fkelava, line 147: This doesn't help. You can still get the directory deleted between this check and `File.Exists(manifest_path)` below. Just ditch this and take the fault at the point you try to read the manifest. "Manifest not found" does not need to clarify whether that is because the file isn't there or the entire directory was torched.

fkelava (changes requested):

src/modmgr/ui_actions.cs
```diff
@@ -0,0 +1,213 @@
...
+        if (pending == null) {
+            action = null!;
+            return false;
+        }
+
+        action = pending;
+        pending = null;
+
+        return true;
+    }
```
fkelava, line 137: You can do away with this and replace all the usages with a snippet roughly as follows (adapt to the actual target type) - don't forget to add a `global using System.Threading` in `usings.cs`;
```cs
FhPendingModToggle? action = Interlocked.Exchange(ref _pending_mod_toggle, null);
if (action == null)
    return;
```

src/modmgr/mods.cs
```diff
@@ -0,0 +1,394 @@
...
+            JsonElement root = document.RootElement;
+
+            // Version/Authors are purely cosmetic, so they default to blank and get
+            // skipped by the UI; Name falls back to the mod_id
+            string name     = _read_string(root, "Name", mod_id);
+            string version  = _read_string(root, "Version", "0");
+            string authors  = _read_string(root, "Authors", "Unknown");
+
+            return new(new FhManifest(mod_id, name, "", authors, version, "", [], [], FhManifestFlags.NONE), mod_directory, true, true, "", load_order_index);
+        }
```
fkelava, line 167: I'm perplexed by why this is not simply:
```cs
try {
    FhManifest manifest = JsonSerializer.Deserialize<FhManifest>(File.ReadAllText(manifest_path))
        ?? throw new Exception($"Manifest invalid or illegible: {manifest_path}");
        
    return new(manifest, // ...
}
// ... catch-block as it is
```
You can then override empty fields if you really care to. But I don't follow why we're creating a manifest from whole cloth when it's _right there_.

fkelava (changes requested):

src/modmgr/mods.cs
```diff
@@ -0,0 +1,394 @@
...
+
+        if (property.ValueKind != JsonValueKind.String) {
+            return fallback;
+        }
+
+        string? readValue = property.GetString();
+        return string.IsNullOrEmpty(readValue) ? fallback : readValue!;
+    }
+}
```
fkelava, line 188: Considering the last comment about manifest deserialization, this can be removed.

src/modmgr/settings.cs
```diff
@@ -0,0 +1,105 @@
...
+            File.WriteAllText(temporary_path, contents);
+            File.Move(temporary_path, destination, true);
+        }
+        finally {
+            if (File.Exists(temporary_path)) {
+                File.Delete(temporary_path);
+            }
+        }
+    }
+}
```
fkelava, line 105: This doesn't seem right. For one, this `finally` block itself can throw. Then what?

If the intended contract here is to throw on failure, then you can reduce the method to this:
```cs
string temporary_path = $"{destination}.tmp";
string backup_path    = $"{destination}.bak";

File.WriteAllText(temporary_path, contents);
File.Replace     (temporary_path, destination, backup_path);
```
`File.Replace` is the actually atomic way to do what you wanted[^1]. As a bonus you get a backup copy.

If you want to handle throws, then wrap the latter two calls in a `try` with a non-throwing `catch`.

[^1]: https://zenn.dev/arika/articles/20251010-csharp-replace-is-atomic?locale=en

fkelava (changes requested):

src/modmgr/mods.cs
```diff
@@ -0,0 +1,394 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit.Tools.ModManager;
+
+
```
fkelava, line 8: Trim instances of double empty lines anywhere they might be found to single empty lines, ex. this, L29-30 in the same file, etc.

fkelava (changes requested):

PR #235, opened 2026-07-27
Enable Excel file support and prepare EEdit stubs

### LINKED ISSUE #121: `ExcelFile` causes `TypeLoadException`, breaking some build-time flows
For instance, one cannot use [Disasmo](https://github.com/EgorBo/Disasmo) to compare JIT output for various methods because it'll throw a:
```
Unhandled exception. System.Reflection.ReflectionTypeLoadException: Unable to load one or more of the requested types.
Could not load type 'Fahrenheit.FFX.ExcelFile`1' from assembly 'fh, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' because generic types cannot have explicit layout.
   at System.Reflection.RuntimeModule.GetDefinedTypes()
   at System.Reflection.RuntimeModule.GetTypes()
   at DisasmoLoader.PrecompileAllMethodsInType(String[] args)
   at DisasmoLoader.Main(String[] args)
```

We should fix this before the next release.

### LINKED ISSUE #234: Model the remaining `kernel` Excel types
Part and parcel of the Excel/EEdit work for `alpha12` is modeling more Excel types so the editor can properly interpret them. We should at least go through all the `kernel` types that are known.

fkelava:
Fixes #121, resolves #234.

One way in which Fahrenheit is lacking compared to the state-of-the-art tooling is absence of functional support for manipulating Excel files. Excel file modding is now considered ubiquitous and mundane, and new tools allowing for it have proliferated rapidly after the advent of the [FFXDataParser](https://github.com/Karifean/FFXDataParser).

While these tools are fit for purpose, they are all fundamentally limited. With `FhEncoding` and DEdit as its frontend, Fahrenheit demonstrated a better option: an API that can power the same transformations in and out of game.

It's time to do the same for Excel files. `alpha12` should feature an initial iteration of Excel interop and its frontend, EEdit.


src/core/ffx/excel.cs
```diff
@@ -65,124 +65,61 @@ internal struct ExcelHeader {
...
-
-        return null;
+        return MemoryMarshal.TryRead(_bytes, out ExcelFileProlog<T> prolog)
+            ? MemoryMarshal.Cast<byte, ExcelHeader>(_bytes [ sz_prolog .. (sz_prolog + prolog.header_count * sz_header) ])
+            : [];
     }
 
     /// <summary>
-    ///     Gets the element at the specified index, with boundary checking.
+    ///     The elements defined in the given <paramref name="header"/>.
```
EvelynTSMG: This is a method, and not a property. Therefore, its primary purpose is to *do* something. The documentation should start with a verb.

The current documentation is more fit for the `<returns>` block.

EvelynTSMG: This documentation is missing both a `<param>` block for the `header` parameter and a `<returns>` block.

src/core/ffx/excel.cs
```diff
@@ -65,124 +65,61 @@ internal struct ExcelHeader {
...
-        ExcelHeader header = find_header_for_index(index)
-            ?? throw new IndexOutOfRangeException($"Cannot find header for index {index} in excel file.");
-
-        return get_text(header, offset);
+        return MemoryMarshal.Cast<byte, T>(_bytes [ start .. (start + length) ]);
     }
+}
 
+/// <summary>
+///     A container for an array of <typeparamref name="T"/>.
```
EvelynTSMG: This is the prologue; you can't even get the headers out of it, so it can hardly be called a container from the user's perspective.

src/core/ffx/excel.cs
```diff
@@ -65,124 +65,61 @@ internal struct ExcelHeader {
...
+///     These files make up most of the game's <c>kernel</c> folder.
```
EvelynTSMG: This sentence would currently better fit as either part of the `ExcelFileReader<T>` documentation or somewhere at the top of the file in an explanation of the Excel system as a whole.

src/core/ffx/excel.cs
EvelynTSMG: Maybe like this?
```suggestion
///     The prologue of an Excel file, which is a container for an array of <typeparamref name="T"/>.
```

src/core/ffx/excel.cs
EvelynTSMG: If the above suggestion is applied, this comment is no longer valid.

src/eedit/editors.cs
```diff
@@ -6,24 +6,41 @@
...
+        foreach (ExcelHeader header in iter.headers()) {
+            _elements.AddRange(iter.elements(header));
+        }
     }
+
 }
 
-internal unsafe class EditorKaizou : EEditComponent<CustomizationRecipe> {
-    internal override void render() {
+// etc. - insert actual UI per wishes
```
EvelynTSMG: Leave TODO(s) instead of this comment.

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
 ///     The well-known Excel file types in enum form, used for internal ordering.
```
EvelynTSMG, line 9: Out of curiosity, how would this work if someone made a custom excel file for their mod?

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
 
     /// <summary>
     ///     Obtains a descriptive display string for the <see cref="EEditComponent"/>
     ///     that implements editing the selected <paramref name="mode"/>.
     /// </summary>
     private static string _get_component_name_by_mode(EEditMode mode) {
         return mode switch {
-            EEditMode.KAIZOU     => $"{nameof(CustomizationRecipe)} - (kaizou.bin)",
-            EEditMode.SUM_ASSURE => $"{nameof(AeonStatBoostsScaling)} - (sum_assure.bin)",
+            EEditMode.A_ABILITY  => $"(a_ability.bin)",
```
EvelynTSMG: [`AutoAbility`](https://github.com/fahrenheit-crew/fahrenheit/blob/6120c35245fa71b9ed4ed3fefddccaa247d8e444/src/core/ffx/aability.cs)

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
+            EEditMode.MONMAGIC   => $"(monmagic.bin)",
+            EEditMode.MONMAGIC1  or
+            EEditMode.MONMAGIC2  => $"{nameof(Command)} - (monmagic1,2.bin)",
+            EEditMode.MONSTER1   => $"(monster1.bin)",
+            EEditMode.MONSTER2   => $"(monster2.bin)",
+            EEditMode.MONSTER3   => $"(monster3.bin)",
+            EEditMode.NAME_TXT   => $"(name_txt.bin)",
+            EEditMode.PANEL      => $"(panel.bin)",
+            EEditMode.PARTY      => $"(party.bin)",
+            EEditMode.PLY_ROM    => $"(ply_rom.bin)",
```
EvelynTSMG: [`PlyRom`](https://github.com/fahrenheit-crew/fahrenheit/blob/6120c35245fa71b9ed4ed3fefddccaa247d8e444/src/core/ffx/plyrom.cs)

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
+            EEditMode.PLY_SAVE   => $"(ply_save.bin)",
```
EvelynTSMG: [`PlySave`](https://github.com/fahrenheit-crew/fahrenheit/blob/6120c35245fa71b9ed4ed3fefddccaa247d8e444/src/core/ffx/plysave.cs), starting state on new game

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
+            EEditMode.CONFIG_TXT => $"(config_txt.bin)",
+            EEditMode.CTB_BASE   => $"(ctb_base.bin)",
+            EEditMode.HELP_TXT   => $"(help_txt.bin)",
+            EEditMode.IMPORTANT  => $"(important.bin)",
+            EEditMode.ITEM       => $"{nameof(PCommand)} - (item.bin)",
+            EEditMode.ITEM_GET   => $"(item_get.bin)",
+            EEditMode.ITEM_RATE  => $"(item_rate.bin)",
+            EEditMode.ITEM_SHOP  => $"(item_shop.bin)",
+            EEditMode.ITEM_TXT   => $"(item_txt.bin)",
+            EEditMode.KAIZOU     => $"(kaizou.bin)",
```
EvelynTSMG: [`CustomizationRecipe`](https://github.com/fahrenheit-crew/fahrenheit/blob/6120c35245fa71b9ed4ed3fefddccaa247d8e444/src/core/ffx/aability.cs#L149)

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
+            EEditMode.PLY_ROM    => $"(ply_rom.bin)",
+            EEditMode.PLY_SAVE   => $"(ply_save.bin)",
+            EEditMode.PREPARE    => $"(prepare.bin)",
+            EEditMode.SAVE_TXT   => $"(save_txt.bin)",
+            EEditMode.SHOP_ARMS  => $"(shop_arms.bin)",
+            EEditMode.SPHERE     => $"(sphere.bin)",
+            EEditMode.ST_NUMBER  => $"(st_number.bin)",
+            EEditMode.STATUS_TXT => $"(status_txt.bin)",
+            EEditMode.SUM_ASSURE => $"(sum_assure.bin)",
+            EEditMode.SUM_GROW   => $"(sum_grow.bin)",
```
EvelynTSMG: `AeonStatBoostsMinimum` and `AeonAbilityRecipe` from [summon.cs](https://github.com/fahrenheit-crew/fahrenheit/blob/6120c35245fa71b9ed4ed3fefddccaa247d8e444/src/core/ffx/summon.cs), respectively.

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
     private static string _get_component_name_by_mode(EEditMode mode) {
         return mode switch {
-            EEditMode.KAIZOU     => $"{nameof(CustomizationRecipe)} - (kaizou.bin)",
-            EEditMode.SUM_ASSURE => $"{nameof(AeonStatBoostsScaling)} - (sum_assure.bin)",
+            EEditMode.A_ABILITY  => $"(a_ability.bin)",
+            EEditMode.AMAPDATA   => $"(amapdata.bin)",
+            EEditMode.ARMS_RATE  => $"(arms_rate.bin)",
+            EEditMode.ARMS_SHOP  => $"(arms_shop.bin)",
+            EEditMode.ARMS_TXT   => $"(arms_txt.bin)",
+            EEditMode.BTL        => $"(btl.bin)",
```
EvelynTSMG: This isn't an excel file, it's a chunked file.

src/eedit/ui_modeselect.cs
```diff
@@ -33,8 +133,58 @@ private static string _get_component_name_by_mode(EEditMode mode) {
...
+            EEditMode.SHOP_ARMS  => throw new NotImplementedException("#234"),
+            EEditMode.SPHERE     => throw new NotImplementedException("#234"),
+            EEditMode.ST_NUMBER  => throw new NotImplementedException("#234"),
+            EEditMode.STATUS_TXT => throw new NotImplementedException("#234"),
+            EEditMode.SUM_ASSURE => throw new NotImplementedException("#234"),
+            EEditMode.SUM_GROW   => throw new NotImplementedException("#234"),
+            EEditMode.SUMMON_TXT => throw new NotImplementedException("#234"),
+            EEditMode.TAKARA     => throw new NotImplementedException("#234"),
+            EEditMode.W_NAME     => throw new NotImplementedException("#234"),
+            EEditMode.WEAPON     => throw new NotImplementedException("#234"),
```
EvelynTSMG: Here, a better message you can copy paste:
```
"Not yet implemented. See fahrenheit-crew/fahrenheit#234 over on GitHub for more details: https://github.com/fahrenheit-crew/fahrenheit/issues/234"
```

Alternatively, just remove the message. It's not incredibly useful, and the NotImplementedException will either be removed or need a different message before this PR is merged anyway.

src/eedit/ui_modeselect.cs
```diff
@@ -9,21 +9,121 @@ namespace Fahrenheit.Tools.EEdit;
...
+            EEditMode.CTB_BASE   => $"(ctb_base.bin)",
+            EEditMode.HELP_TXT   => $"(help_txt.bin)",
+            EEditMode.IMPORTANT  => $"(important.bin)",
+            EEditMode.ITEM       => $"{nameof(PCommand)} - (item.bin)",
+            EEditMode.ITEM_GET   => $"(item_get.bin)",
+            EEditMode.ITEM_RATE  => $"(item_rate.bin)",
+            EEditMode.ITEM_SHOP  => $"(item_shop.bin)",
+            EEditMode.ITEM_TXT   => $"(item_txt.bin)",
+            EEditMode.KAIZOU     => $"(kaizou.bin)",
+            EEditMode.MAGIC      => $"(magic.bin)",
```
EvelynTSMG: This isn't an excel file.

EvelynTSMG (changes requested):
Mostly documentation changes. The amount of comments is a bit exaggerated, as I tend to write one comment per *instance* of an issue, not one per issue.

src/eedit/ui_modeselect.cs
fkelava: Probably `EEditMode.GENERIC`, `EditorGeneric<T>`, and a boatload of reflection.

src/eedit/ui_modeselect.cs
```diff
@@ -32,36 +31,32 @@ internal enum EEditMode {
...
+    SPHERE      = 39,
+    ST_NUMBER   = 40,
+    STATUS_TXT  = 41,
+    SUM_ASSURE  = 42,
+    SUM_GROW    = 43,
+    SUMMON_TXT  = 44,
+    TAKARA      = 45,
+    W_NAME      = 46,
+    WEAPON      = 47,
+    COUNT_TYPES = 48
```
EvelynTSMG: Put a blank line between these so it's easier to recognize `COUNT_TYPES` as not being an excel type

EvelynTSMG (changes requested):
Just an extra note I didn't realize earlier, sorry

src/core/ffx/excel.cs
```diff
@@ -18,171 +26,134 @@ public struct ExcelTextOffset {
...
     ///     The offset to the standard text.
     /// </summary>
     public ExcelTextOffset standard;
 
     /// <summary>
-    ///     The offset to the simplified text.<br/>
-    ///     In Japanese, this would have been the hiragana version of the text.<br/>
+    ///     The offset to the simplified text. In Japanese, this is hiragana;
+    ///     in Western encodings, it has no effect.
+    ///     <para/>
```
EvelynTSMG: "Would have been" was correct. This is completely unused; even in the data files I've seen, the simplified text contains a placeholder. The only way we know it would be hiragana is from internal function names using "Hira" for it.

src/core/ffx/important.cs
```diff
@@ -0,0 +1,16 @@
...
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit.FFX;
+
+[StructLayout(LayoutKind.Sequential)]
+public struct KeyItem {
+    public ExcelSimplifiableTextOffset name;
+    public ExcelSimplifiableTextOffset help;
+    public byte                        item_type;  // in FFX, "is Al Bhed Primer"?
```
EvelynTSMG: This is in the FFX namespace, why make this comment instead of naming it `is_primer`?

src/core/ffx/sphere.cs
```diff
@@ -0,0 +1,14 @@
...
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit.FFX;
+
+[StructLayout(LayoutKind.Sequential)]
+public struct SphereUsage {
+    public ExcelSimplifiableTextOffset help;
+    public ushort                      type;
+    public ushort                      activates;
```
EvelynTSMG: This should be a `[Flags]` enum. I believe the ImHex pattern you have should contain a correct version of it.

src/core/ffx/sphere.cs
```diff
@@ -0,0 +1,14 @@
...
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit.FFX;
+
+[StructLayout(LayoutKind.Sequential)]
+public struct SphereUsage {
+    public ExcelSimplifiableTextOffset help;
+    public ushort                      type;
```
EvelynTSMG: This should be an enum. I believe the ImHex pattern you have may contain a correct version of it.

EvelynTSMG (changes requested):

src/core/ffx/excel.cs
```diff
@@ -6,17 +6,17 @@
 namespace Fahrenheit.FFX;
 
 /* [fkelava 28/07/26 02:36]
- * The game stores various data in 'Excel' files, containers for an array of objects.
- * It is a form of binary serialization. These types model generic Excel containers.
+ * The game stores various data in 'Excel' containers, a form of binary serialization.
+ * One or multiple headers precede an array of items, with optional game-encoded text following them.
```
EvelynTSMG: This statement makes it seem like the order is Headers - Elements - Text,
rather than the accurate Headers - Elements A - Text A - Elements B - Text B.
```suggestion
 * One or multiple headers define sections made of an array of elements and, optionally, text.
```
(Note that the name "header" already implies it precedes something.)

src/core/ffx/excel.cs
```diff
@@ -26,51 +26,54 @@ public struct ExcelTextOffset {
...
 
 /// <summary>
-///     The header of an Excel file.
+///     The header of an Excel container.
 /// </summary>
 [StructLayout(LayoutKind.Sequential)]
 public struct ExcelHeader {
     /// <summary>
-    ///     The index of the first element in the file.
+    ///     The index of the first element in the container.
```
EvelynTSMG: It's worth noting that "container" is inaccurate in this context. Excel headers, while near the beginning of the container, define sections. Unlike what this sentence implies, not every header will share the same `index_first` value in a container with multiple headers.
```suggestion
    ///     The index of the first element in the section.
```

src/core/ffx/excel.cs
```diff
@@ -26,51 +26,54 @@ public struct ExcelTextOffset {
...
+    ///     have been hiragana; in Western encodings, it has no effect.
+    ///     <para/>
     ///     This is completely unused.
     /// </summary>
     internal ExcelTextOffset simplified;
 }
 
 /// <summary>
-///     The header of an Excel file.
+///     The header of an Excel container.
```
EvelynTSMG: While this is technically correct, it leaves out a crucial detail: headers each define a section of the container.
```suggestion
///     The header that defines a section in an Excel container.
```

src/core/ffx/excel.cs
```diff
@@ -26,51 +26,54 @@ public struct ExcelTextOffset {
...
 public struct ExcelHeader {
     /// <summary>
-    ///     The index of the first element in the file.
+    ///     The index of the first element in the container.
     /// </summary>
     public ushort index_first;
 
     /// <summary>
-    ///     The index of the last element in the file.
+    ///     The index of the last element in the container.
```
EvelynTSMG: It's worth noting that "container" is inaccurate in this context. Excel headers, while near the beginning of the container, define sections. Unlike what this sentence implies, not every header will share the same `index_last` value in a container with multiple headers.
```suggestion
    ///     The index of the last element in the section.
```

src/core/ffx/excel.cs
```diff
@@ -26,51 +26,54 @@ public struct ExcelTextOffset {
...
     /// </summary>
     public ushort index_last;
 
     /// <summary>
-    ///     The size of one element in the file.
+    ///     The size of one element in the container.
```
EvelynTSMG: It's worth noting that "container" is inaccurate in this context. Excel headers, while near the beginning of the container, define sections.
```suggestion
    ///     The size of one element in the section.
```

src/core/ffx/excel.cs
```diff
@@ -26,51 +26,54 @@ public struct ExcelTextOffset {
...
     /// </summary>
     public ushort element_size;
 
     /// <summary>
-    ///     The combined length, in bytes, of all the elements in the file.<br/>
-    ///     Beyond the data there may be text, referenced using offsets in the elements.
+    ///     The combined length, in bytes, of all the elements in the container.
```
EvelynTSMG: It's worth noting that "container" is inaccurate in this context. Excel headers, while near the beginning of the container, define sections. Unlike what this sentence implies, not every header will share the same `data_length` value in a container with multiple headers.
```suggestion
    ///     The combined length, in bytes, of all elements in the section.
```
("all elements" vs "all the elements" vs "all of the elements" is personal preference, do whatever)

src/core/ffx/excel.cs
```diff
@@ -127,23 +130,23 @@ private bool find_header_for_index(int index, out ExcelHeader target_header) {
...
 /// <summary>
-///     The prologue of an Excel file. Describes the file in enough detail to construct a reader.
+///     The prologue of an Excel container. Describes it in enough detail to construct a reader.
 ///     <para/>
-///     To iterate over the file's contents, use an <see cref="ExcelFileReader{T}"/>.
+///     To iterate over its contents, use an <see cref="ExcelReader{T}"/>.
 /// </summary>
 [StructLayout(LayoutKind.Sequential, Size = 0x8)]
-public struct ExcelFileProlog {
+public struct ExcelProlog {
```
EvelynTSMG: It is weird to me that this structure is only defined *after* `ExcelReader` despite the other structs it depends on, such as `ExcelHeader`, being defined before it.

src/core/ffx/sphere.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace Fahrenheit.FFX;
+
+public enum SphereBehavior : ushort {
+    NONE      = 0,
+    ACTIVATOR = 1,
+    MODIFIER  = 2
+}
+
+[Flags]
+public enum SphereApplicability : ushort {
```
EvelynTSMG: Could we name this `SphereTargets`? It seems like less of a mouthful.

src/core/ffx/sphere.cs
```diff
@@ -0,0 +1,35 @@
...
+    MP            = 1 << 9,
+    ABILITY       = 1 << 10
+}
+
+[StructLayout(LayoutKind.Sequential)]
+public struct Sphere {
+    public ExcelSimplifiableTextOffset help;
+    public SphereBehavior              type;
+    public SphereApplicability         activates;
+    public int                         _0x0C;
```
EvelynTSMG: For aesthetic's sake, please put an empty line between `help` and the other fields.

src/core/ffx/excel.cs
EvelynTSMG: This is absolutely my fault, for which I apologize, but I would be amiss to not mention that the `remarks` tag is supposed to be separate from the `summary` tag, as shown in the example over in the [C# language specification for documentation comments](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d312-remarks).

EvelynTSMG (changes requested):
Mostly small documentation changes

src/core/ffx/excel.cs
```diff
@@ -5,184 +5,155 @@
...
-public struct ExcelTextOffset {
-    /// <summary>
-    ///     The offset to the text.
-    /// </summary>
-    public  ushort text_offset;
-    private ushort __0x2; // Clearly related to the text, but unknown
-}
+/* [fkelava 28/07/26 02:36]
+ * The game stores various data in 'Excel' containers, a form of binary serialization.
+ * 'Headers' define 'sections' consisting of an array of elements and, optionally, text.
```
EvelynTSMG: Why the quotes????

fkelava: Simply as a crude form of emphasis. I can delete them if they bother you.

EvelynTSMG: That's. not how emphasis works? They're more confusing than anything here. I believe they should be deleted.

src/core/ffx/excel.cs
```diff
@@ -5,184 +5,155 @@
...
-    ///     Beyond the data there may be text, referenced using offsets in the elements.
+    ///     The combined length, in bytes, of all the elements in the section.
+    ///     <para/>
+    ///     This does not include any text which may follow the data.
     /// </summary>
     public ushort data_length;
 
     /// <summary>
-    ///     An offset from the start of the file to the start of the data.<br/>
+    ///     The offset, in bytes, from the start of the section to the start of the data.
```
EvelynTSMG: While I didn't notice the previous version should've had "file" replaced with "container", there's a reason I didn't put a comment here in my last review.

This offset *is*, in fact, from the start of the *container*, not the specific section.
```suggestion
    ///     The offset from the start of the container to the start of the section.
```
(Also, when are offsets in anything but bytes?)

src/core/ffx/excel.cs
```diff
@@ -5,184 +5,155 @@
...
+    /// </remarks>
+    public  ushort header_count;
+    private ushort _0x02;
+    private ushort _0x04;
+    private ushort _0x06;
 }
 
 /// <summary>
-///     The header of Excel files.
+///     The header of an Excel section.
```
EvelynTSMG: This now implies headers appear at the beginning of each section. Let's get rid of the "of" altogether.
```suggestion
///     The header that defines a section in an Excel container.
```

EvelynTSMG (changes requested):

src/core/ffx/excel.cs
fkelava: The `byte` thing is to avoid the common pitfall of trying to add an offset to a `T*`, which will result in adding `offset * sizeof(T)` bytes. It implies that the other addend must be cast to `byte*`.

fkelava:
The last series of commits should equalize the definitions of any objects I've bothered to fix between the data parser and Fahrenheit. Emphasis _should_. This should be double-checked.

fkelava:
A note: `Size` on `StructLayout.Sequential` structs has no effect. This is why it has been removed where applicable.

(Yes, I know we've done otherwise for a long time; yes, I'm gonna fix that in a followup.)

src/core/ffx/customize.cs
EvelynTSMG, line 1: We named the MixRecipe file `mix.cs` instead of `prepare.cs`. Similarly, we should name this `customize.cs` or similar instead of `kaizou.cs`.

src/core/ffx/aability.cs
```diff
@@ -5,33 +5,33 @@
...
-    [FieldOffset(0x60)] public StatusExtraFlags      status_resist_extra;
-    [FieldOffset(0x62)] public AutoAbilityEffectsMap auto_ability_effects;
-    [FieldOffset(0x68)] public byte                  icon;
-    [FieldOffset(0x69)] public byte                  group_idx;
-    [FieldOffset(0x6A)] public byte                  group_level;
-    [FieldOffset(0x6B)] public byte                  international_bonus_idx;
+    public ExcelTextOffset       name_offset;
+    public ExcelTextOffset       dash_offset;
+    public ExcelTextOffset       desc_offset;
+    public ExcelTextOffset       misc_offset;
```
EvelynTSMG: On a closer look at the file, these appear to be two simplifiable offsets:
```suggestion
    public ExcelSimplifiableTextOffset name;
    public ExcelSimplifiableTextOffset desc;

```
(Note that as you did above in this PR, the `_offset` should be removed in this case)

src/core/ffx/aability.cs
```diff
@@ -5,33 +5,33 @@
...
+    public StatusPermanentFlags  status_auto_permanent;
+    public StatusTemporalFlags   status_auto_temporal;
+    public StatusExtraFlags      status_auto_extra;
+    public StatusExtraFlags      status_inflict_extra;
+    public StatusExtraFlags      status_resist_extra;
+    public AutoAbilityEffectsMap auto_ability_effects;
+    public byte                  icon;
+    public byte                  group_idx;
+    public byte                  group_level;
+    public byte                  international_bonus_idx;
```
EvelynTSMG: Let's add some blank lines for readability, shall we?
I recommend adding them between:
- `desc`, `is_sos`, and `elem_strike`
- `elem_weak` and `status_inflict`
- `status_resist` and `stat_inc_amount`,
- `stat_inc_flags` and `status_auto_permanent`,
- `status_auto_extra` and `status_inflict_extra`,
- `status_resist_extra`, `auto_ability_effects`, and `icon`

src/core/ffx/command.cs
```diff
@@ -5,55 +5,56 @@
...
-    [FieldOffset(0x47)] public  StatusDurationMap status_duration_map;
-    [FieldOffset(0x54)] public  StatusExtraFlags  flags_status_extra;
-    [FieldOffset(0x56)] public  byte              flags_buffs_stat;
-    [FieldOffset(0x58)] public  byte              overdrive_category;
-    [FieldOffset(0x59)] public  byte              buff_amount;
-    [FieldOffset(0x5A)] public  byte              flags_buffs_mix;
+    public ExcelTextOffset   name_offset;
+    public ExcelTextOffset   dash_offset;
+    public ExcelTextOffset   desc_offset;
+    public ExcelTextOffset   misc_offset;
```
EvelynTSMG: On a closer look at the file, these appear to be two simplifiable offsets:
```suggestion
    public ExcelSimplifiableTextOffset name;
    public ExcelSimplifiableTextOffset desc;

```
(Note that as you did above in this PR, the `_offset` should be removed in this case)

src/core/ffx/command.cs
```diff
@@ -5,55 +5,56 @@
...
+    public byte              hit_count;
+    public byte              shatter_chance;
+    public ElementFlags      element;
+    public StatusMap         status_map;
+    public StatusDurationMap status_duration_map;
+    public StatusExtraFlags  flags_status_extra;
+    public ushort            flags_buffs_stat;
+    public byte              overdrive_category;
+    public byte              buff_amount;
+    public ushort            flags_buffs_mix;
```
EvelynTSMG: Let's add some blank lines for readability, shall we?
I recommend adding them between:
- `desc` and `anim1`
- `caster_anim` and `flags_menu`
- `user_id` and `flags_target`
- (optionally) `steals_gil` and `party_preview`
- `flags_damage_class` and `ctb_rank`
- `limit_cost` and `crit_bonus`
- `shatter_chance` and `element`
- `element` and `status_map`
- `flags_status_extra` and `flags_buff_stat`

src/core/ffx/plysave.cs
```diff
@@ -10,53 +10,53 @@ public struct PlySaveLimitModeCtrArray {
...
-    [FieldOffset(0x4A)] public  AutoAbilityEffectsMap    auto_ability_effects;
-    [FieldOffset(0x50)] public  uint                     battle_count;
-    [FieldOffset(0x54)] public  uint                     enemies_defeated;
-    [FieldOffset(0x58)] public  uint                     deaths;
-    [FieldOffset(0x5C)] public  uint                     limits_charged;
-    [FieldOffset(0x60)] public  PlySaveLimitModeCtrArray limit_mode_counters;
-    [FieldOffset(0x88)] public  OverdriveModeFlags       obtained_limit_modes;
-    [FieldOffset(0x8C)] private uint                     __0x8C;
-    [FieldOffset(0x90)] private uint                     __0x90;
+    public  ExcelTextOffset          name_offset;
```
EvelynTSMG: This field should be called `name` instead of `name_offset`.

src/core/ffx/plysave.cs
```diff
@@ -10,53 +10,53 @@ public struct PlySaveLimitModeCtrArray {
...
+    public  AbilityMap               abi_map;
+    public  AutoAbilityEffectsMap    auto_ability_effects;
+    public  uint                     battle_count;
+    public  uint                     enemies_defeated;
+    public  uint                     deaths;
+    public  uint                     limits_charged;
+    public  PlySaveLimitModeCtrArray limit_mode_counters;
+    public  OverdriveModeFlags       obtained_limit_modes;
+    private uint                     __0x8C;
+    private uint                     __0x90;
```
EvelynTSMG: Let's add some blank lines for readability, shall we?
I recommend adding them between:
- `name` and `base_hp`
- `base_accuracy` and `total_ap`
- `ap` and `hp`
- `max_mp` and `ply_flags`
- `ply_flags` and `wpn_inv_idx`
- `arm_inv_idx` and `strength`
- `accuracy and `poison_dmg`
- `poison_dmg` and `limit_mode_index`
- `limit_charge_max` and `slv_available`
- `slv_spent` and `battles_until_recovery`
- `battles_until_recovery` and `abi_map`
- `auto_ability_effects` and `battle_count`
- `limit_mode_coutnters` and `obtained_limit_modes`
- `obtained_limit_modes` and `__0x8C`

src/core/ffx/shop.cs
```diff
@@ -0,0 +1,22 @@
...
+
+[StructLayout(LayoutKind.Sequential)]
+public struct Shop {
+
+    [InlineArray(0x10)]
+    public struct ShopOffers {
+        private ushort _e0;
+    }
+
+    private ushort     _0x00;
```
EvelynTSMG: This is a percentage that multiplies the price, usually 100 for the regular price. One shop in the game appears to use 150 for x1.5 the regular price.
```cs
    /// <summary>The percentage the base price of each offer should be multiplied by.</summary>
    public ushort price_multiplier;
```

src/core/ffx/sphere.cs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+[StructLayout(LayoutKind.Sequential)]
+public struct Sphere {
+    public  ExcelSimplifiableTextOffset help;
+    public  SphereBehavior              type;
+    public  SphereTargets               activates;
+    public  SphereRange                 range;
+    public  byte                        special_role;
+    private ushort                      _0x0E;
```
EvelynTSMG: Let's add some blank lines for readability, shall we?
I recommend adding them between:
- `help` and `type`
- `special_role` and `_0x0E`

src/eedit/ui_modeselect.cs
```diff
@@ -9,22 +9,162 @@ namespace Fahrenheit.Tools.EEdit;
 ///     The well-known Excel file types in enum form, used for internal ordering.
 /// </summary>
 internal enum EEditMode {
-    KAIZOU           = 0,
-    SUM_ASSURE       = 1,
-    FILE_TYPES_COUNT = 2
+    // Common
+    NULL = 1,
+    TXT  = 2,
+    RATE = 3,
```
EvelynTSMG: `*_rate.bin` are not some generic type, they define specific values (item / gear prices), so they should be in the lower group alongside `ITEM_SHOP` etc.

src/eedit/ui_modeselect.cs
```diff
@@ -9,22 +9,162 @@ namespace Fahrenheit.Tools.EEdit;
...
 
     /// <summary>
     ///     Obtains a descriptive display string for the <see cref="EEditComponent"/>
     ///     that implements editing the selected <paramref name="mode"/>.
     /// </summary>
     private static string _get_component_name_by_mode(EEditMode mode) {
         return mode switch {
+            EEditMode.NULL       => $"(Unknown Type)",
+            EEditMode.TXT        => $"{nameof(NameHelpText)} - (*_txt.bin)",
+            EEditMode.RATE       => $"{nameof(Rate)} - (*_rate.bin)",
```
EvelynTSMG: `*_rate.bin` are not some generic type, they define specific values (item / gear prices), so they should be in the lower group alongside `ITEM_SHOP` etc.

src/eedit/ui_modeselect.cs
```diff
@@ -33,75 +173,177 @@ private static string _get_component_name_by_mode(EEditMode mode) {
     /// </summary>
     private static EEditComponent _get_component_by_mode(EEditMode mode) {
         return mode switch {
+            EEditMode.TXT        => new EditorTextPair(),
+            EEditMode.RATE       => new EditorRate(),
```
EvelynTSMG: `*_rate.bin` are not some generic type, they define specific values (item / gear prices), so they should be in the lower group alongside `ITEM_SHOP` etc.

src/eedit/ui_modeselect.cs
```diff
@@ -33,75 +173,177 @@ private static string _get_component_name_by_mode(EEditMode mode) {
...
+            "help_txt.bin"   or
+            "item_txt.bin"   or
+            "menu_txt.bin"   or
+            "menu_txt2.bin"  or
+            "mmain_txt.bin"  or
+            "name_txt.bin"   or
+            "save_txt.bin"   or
+            "status_txt.bin" or
+            "summon_txt.bin" => EEditMode.TXT,
+            "arms_rate.bin"  or
```
EvelynTSMG, line 275: Any groupings of multiple file names with "or" should probably be surrounded by blank lines.

test/core/excel.cs
```diff
@@ -0,0 +1,38 @@
...
+[TestFixture]
+public class FhExcelTests {
+
+    [Test]
+    public void ffx_excel_sizeof() {
+        using (Assert.EnterMultipleScope()) {
+            Assert.That(Unsafe.SizeOf<FFX.AutoAbility>(),         Is.EqualTo(0x6C)); // a_ability.bin
+            Assert.That(Unsafe.SizeOf<FFX.Command>(),             Is.EqualTo(0x5C)); // {item|command|monmagic*}.bin
+            Assert.That(Unsafe.SizeOf<FFX.PCommandData>(),        Is.EqualTo(0x04)); // ^
+            Assert.That(Unsafe.SizeOf<FFX.PCommand>(),            Is.EqualTo(0x60)); // ^
```
EvelynTSMG: `Command` technically only `monmagic{1|2}.bin`, and `PCommand` is in `{item|command}.bin`.
```suggestion
            Assert.That(Unsafe.SizeOf<FFX.Command>(),             Is.EqualTo(0x5C)); // monmagic{1|2}.bin
            Assert.That(Unsafe.SizeOf<FFX.PCommandData>(),        Is.EqualTo(0x04)); // PCommand
            Assert.That(Unsafe.SizeOf<FFX.PCommand>(),            Is.EqualTo(0x60)); // {item|command}.bin
```

EvelynTSMG (changes requested):

src/core/ffx/plysave.cs
```diff
@@ -12,49 +12,63 @@ public struct PlySaveLimitModeCtrArray {
...
+    public AbilityMap            abi_map;
+    public AutoAbilityEffectsMap auto_ability_effects;
+
+    public uint                     battle_count;
+    public uint                     enemies_defeated;
+    public uint                     deaths;
+    public uint                     limits_charged;
+    public PlySaveLimitModeCtrArray limit_mode_counters;
+
+    public OverdriveModeFlags obtained_limit_modes;
```
EvelynTSMG: The idea behind my initial request was to group:
- the stats: `battle_count`, `enemies_defeated`, `deaths`, and `limits_charged`
- the limit (mode) related: `limit_mode_counters` and `obtained_limit_modes`

I understand that `limits_charged` looks like it belongs in the limit-related group, but it actually belongs in the stats group.

<insert Calvin's dad "there are two types" meme here>

src/eedit/ui_modeselect.cs
```diff
@@ -94,24 +94,24 @@ internal static partial class UI {
...
+            EEditMode.ITEM_GET  => $"(item_get.bin)",
+            EEditMode.ITEM_SHOP => $"(item_shop.bin)",
+            EEditMode.MONMAGIC  => $"(monmagic.bin)",
+            EEditMode.MONSTER   => $"(monster*.bin)",
+            EEditMode.PARTY     => $"(party.bin)",
+            EEditMode.PLY_ROM   => $"{nameof(PlyRom)} - (ply_rom.bin)",
+            EEditMode.PLY_SAVE  => $"{nameof(PlySave)} - (ply_save.bin)",
+            EEditMode.PREPARE   => $"{nameof(MixRecipe)} - (prepare.bin)",
+            EEditMode.ST_NUMBER => $"{nameof(StNumber)} - (st_number.bin)",
+            EEditMode.TAKARA    => $"{nameof(Treasure)} - (takara.bin)",
```
EvelynTSMG: These are arranged alphabetically, why is `RATE` at the top?  It should be below `PREPARE`.

src/eedit/ui_modeselect.cs
EvelynTSMG: These are arranged alphabetically, why is `RATE` at the top?  It should be below `PREPARE`.

src/eedit/ui_modeselect.cs
```diff
@@ -173,23 +173,23 @@ private static string _get_component_name_by_mode(EEditMode mode) {
...
+            EEditMode.ITEM_GET  => throw new NotImplementedException(),
+            EEditMode.ITEM_SHOP => throw new NotImplementedException(),
+            EEditMode.MONMAGIC  => throw new NotImplementedException(),
+            EEditMode.MONSTER   => throw new NotImplementedException(),
+            EEditMode.PARTY     => throw new NotImplementedException(),
+            EEditMode.PLY_ROM   => new EditorPlyRom(),
+            EEditMode.PLY_SAVE  => new EditorPlySave(),
+            EEditMode.PREPARE   => new EditorPrepare(),
+            EEditMode.ST_NUMBER => new EditorStNumber(),
+            EEditMode.TAKARA    => new EditorTakara(),
```
EvelynTSMG: These are arranged alphabetically, why is `RATE` at the top?  It should be below `PREPARE`.

EvelynTSMG (changes requested):

EvelynTSMG (approved):
Great work! Looks clean

PR #238, opened 2026-07-29
Fuse identical function calls between games

### LINKED ISSUE #237: Fuse identical function calls between games
After #152, `FhCall` possesses entries for most game functions. This is a exceptionally useful and desirable feature, but comes at a cost to binary size because every function now has a delegate and method-handle-returning method.

These games are, however, not entirely disjoint. They were compiled by the same compiler and share portions of source. It follows from this that the binaries will contain instruction-for-instruction identical functions. `FhMethodLocation` and `FhMethodHandle` are fully capable of directing a call to such functions to the correct address/offset depending on the actually loaded binary while only declaring one delegate and providing one method-handle-returning method.

It is therefore desirable to fuse identical functions in this manner. Reducing binary size has gains for both developers (less onerous editing of `call_*.g.cs`) and users (a smaller binary loads faster and consumes less memory).

fkelava:
Fixes #237.

Includes some additional STEP features.


src/step/main.cs
```diff
@@ -3,487 +3,243 @@
...
+            fn_x,
+            _load_globals  (path_global_x));
+
+        FhGameSpecificGenerator emitter_x2 = new FhGameSpecificGenerator(
+            Directory.CreateDirectory(path_output_x2),
+            _load_reject   (path_reject_x2),
+            _load_remap    (path_remap_x2),
+            FhGameId.FFX2,
+            _load_fn       (path_fn_x2),
+            _load_globals  (path_global_x2));
```
EvelynTSMG, line 144: Ideally, the game-specific remaps would inherit data from the common remap and only contain game-specific types themselves.

EvelynTSMG (approved):

src/step/main.cs
fkelava: A thing to consider in a future update.

PR #245, opened 2026-08-02
SGEN: Fix license headers, generate `lm_{monster|monmagic|command}`

fkelava:
When we relicensed to the LGPL, I forgot to regenerate the headers. While at it, we can generate a few extra headers we didn't yet have and make the generation script move the IDs where they belong, since ClangSharpPInvokeGenerator always emits them side-by-side with the actual struct.


EvelynTSMG (approved):
Seems good. Most of this is auto-generated so it's not like there is much to review 😛

Issue #151, opened 2026-05-28
Rework the save UI to be more in line with the vanilla games


EvelynTSMG:

DawnVeilwinter:
Complete as of #273 and #256.

PR #246, opened 2026-08-02
Rework save PAL to support new save UI

### LINKED ISSUE #244: The save PAL needs adapting to support the new save UI
The save PAL dates back to the very first save system replacement. In it, we interfaced with the game's Flash-based Iggy UI and created new 'entries' in the Flash save list by hand. This perfectly captured the look and feel of the game, but was insufferably slow, and was thus abandoned. However, the PAL still has a number of methods to support that particular, now-overridden, user interface.

After we merged Phyre interop, we could make the save UI look and feel as the game's own (see #151), but the PAL does not provide the information the runtime implementation would need to show the proper icons and such. The bits that deal with the Flash UI should either be separated or outright removed, and the fields such UI needs should be rechecked and exposed.

### LINKED ISSUE #247: A save's stored language ID should be used for all PAL-related operations
A base game bug that we replicate perfectly is using the active game language to attempt to decode save information. This causes Tidus' name to appear as gibberish. The game normally reacts to that by forcibly renaming the character, which we also replicate.

But a save _has_ the originating language ID stored. There is nothing preventing us from using it to present the stored information accurately in the save/load menu. The save PAL and UI should take advantage.

fkelava:
Closes #244.

### DIFF
--- a/src/core/encoding.cs
+++ b/src/core/encoding.cs
@@ -5,7 +5,7 @@
 
 namespace Fahrenheit;
 
-public enum FhLangId {
+public enum FhLangId : byte {
     Japanese = 0,
     English  = 1,
     French   = 2,
--- a/src/core/save_mgr.cs
+++ b/src/core/save_mgr.cs
@@ -79,9 +79,6 @@ internal sealed class FhSaves {
             }
 
             FhSavePal.pal_get_location   (display_data.header, display_data.location);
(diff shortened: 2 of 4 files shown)


fkelava:
I left this as a draft in case more fields need adding to facilitate #151.

EvelynTSMG (commented):
Looks good, approval pending further feedback as to whether everything needed for #151 was included.

EvelynTSMG (approved):
Looks good, thanks!

PR #254, opened 2026-08-06
Fix the game's memory allocator

### LINKED ISSUE #253: 'Solve' the game's memory allocation quirks once and for all
Probably the most oft-repeated folk wisdom and admonition for the X/X-2 Remasters is the necessity of the '4GB patch'. This simply flips the `IMAGE_FILE_LARGE_ADDRESS_AWARE` bit in the binary's [PE header `Characteristics` field](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#characteristics). This is normally specified with the linker option [/LARGEADDRESSAWARE](https://learn.microsoft.com/en-us/cpp/build/reference/largeaddressaware-handle-large-addresses?view=msvc-170), but for any which reason the game binaries do not have this enabled.

The games, because of this, purportedly 'run out of memory', causing FMV green-screens and/or crashes, which this patch purportedly addresses. The former is _technically_ correct, but the latter is an outright lie.

The game has a reasonably well-behaved primary allocator. It uses [VirtualAlloc](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc) to reserve a pool `0x3000_0000` large early in the boot sequence, and then commits units from it as needed, rounded to the nearest page size. Some other minor pools are allocated, but the primary pool is the source of most memory the game dispenses. See ``FFX.exe+{6eab70|6eb310|6ebd00|6ebd20|6ebdc0|9438b0|9439a0}`` and their X-2 equivalents for details.

It follows from this that _very early on in the boot process_, a contiguous block comprising 37.5% of the default address space is made permanently unavailable. The problem arises because of how this interacts with two other quirks.

_One_, in sufficiently modern Windows (Vista and onward) all DLLs are marked [ASLR-capable](https://learn.microsoft.com/en-us/cpp/build/reference/dynamicbase-use-address-space-layout-randomization?view=msvc-170) by default. The OS loader will therefore map them into the process at a random high address. This is easily inspectable with a tool like [VMMap](https://learn.microsoft.com/en-us/sysinternals/downloads/vmmap).

Image

The attached image demonstrates the problem clearly. Due to ASLR, the game's dependencies have been loosely strewn around the upper reaches of the address space. Because they are not tightly packed, obtaining a large contiguous allocation in that entire space is impossible.

Image

When the game then allocates its primary pool, the remaining space is further constrained. So while the game has enough free memory to satisfy its needs, it doesn't have enough _contiguous memory_ to satisfy particularly large allocations.

And what exactly might be a large allocation not accounted for in the primary pool? FMVs, which are conveniently the single largest files the game will attempt to load in its entirety. There isn't enough room for that, and you get a green screen.

Image

So the 4GB patch is the band-aid one applies to fix this, except that _doesn't change the allocator behavior_. FMVs will work because it'll manage to scrounge up some contiguous space past the now-unlocked boundary, but by and large the game has no idea what to do with this space. It doesn't use it. It doesn't know _how to_. More importantly, _it doesn't need it_.

Image

Fahrenheit can, and will, tweak the game's allocation behavior to amend this and enable things to run stably with or without the patch. _With_ the patch, we can invert the direction we allocate in and go top-down to move the game's entire primary heap past the 2G mark; _without_ it, we can tamper with the pool size to leave enough free room for FMVs. The game will, at worst, allocate a smaller new pool and service its requests in this manner.

fkelava:
Fixes #253.


src/runtime/malloc.cs
```diff
@@ -0,0 +1,141 @@
...
+     * places images from circa 0x5000_0000 to 0x7FFF_FFFF like roadblocks, and the large
+     * reserved pool still congests the low reaches of address space. Can we do better?
+     *
+     * The game eventually defers any action by its primary allocator to VirtualAlloc/VirtualFree.
+     * See https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc,
+     * https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree.
+     *
+     * By passing MEM_TOP_DOWN, VirtualAlloc can be induced to go top-down; that is, start allocating
+     * from 7FFF_FFFF downwards. This applies to the primary pool as well! Effectively, we can move
+     * it (and all other primary allocator actions) above the 2G mark.
```
EvelynTSMG: You say it starts allocating from 0x7FFF_FFFF down, and then talk about it as if it was over the 2 GB mark. I believe adding a `(or 0xFFFF_FFFF with the 4gb patch)` would suffice here?

src/runtime/malloc.cs
```diff
@@ -0,0 +1,141 @@
...
+     */
+
+    [UnmanagedCallConv(CallConvs = [ typeof(CallConvCdecl) ] )]
+    private void* h__VirtualAlloc_Reserve_NA(nuint size) {
+        _logger.Info($"MEM_RESERVE(0x{size:X8})");
+        void* rv = FhCall._VirtualAlloc_Reserve_NA.chain_from(h__VirtualAlloc_Reserve_NA).fnptr!(size);
```
EvelynTSMG: Also, you removed the forced top-down behavior in the same commit in which you added that comment. In my opinion, we should absolutely still force it, at least when the executable is large address aware.

src/runtime/malloc.cs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    [UnmanagedCallConv(CallConvs = [ typeof(CallConvCdecl) ] )]
+    private void* h__VirtualAlloc_ReserveCommit_TopDown_RW(nuint size) {
+        _logger.Info($"MEM_RESERVE+COMMIT(TOP_DOWN, 0x{size:X8})");
+        void* rv = FhCall._VirtualAlloc_ReserveCommit_TopDown_RW.chain_from(h__VirtualAlloc_ReserveCommit_TopDown_RW).fnptr!(size);
+        _logger.Info($"0x{(nint)rv:X8}");
+        _logger.Info($"TOTAL RESERVED: 0x{_reserved:X8}");
+        return rv;
+    }
```
EvelynTSMG, line 143: Please make all of this only log at the debug level. If a mod dev or user wants to see the memory usage, they should use the `//TODO`d UI for it

EvelynTSMG (changes requested):
Due to an abundance of care for my mental health, I will trust you that these particular delegates are correct. Also I'm reviewing this from bed on my phone lmao

EvelynTSMG (approved):

PR #258, opened 2026-08-08
Allow reading from the VBF at runtime

### LINKED ISSUE #257: Allow for opening files from VBF at runtime
When dealing with Phyre interop in #55 and #119, we created a system to load such assets at runtime to avoid the problem of bundling copyrighted assets with mods. The logical next step is to extend this generally to permit loading _any_ file from the VBF.

fkelava:
Closes #257.


EvelynTSMG (dismissed):
`FhVbfFileScope` looks good, I assume all the delegates and structs are fine

EvelynTSMG (approved):

PR #260, opened 2026-08-11
Provide a RAII type to manipulate memory protection

fkelava:
Constants in the game process can be writable or read-only. If read-only, the compiler will emit them into `.rdata` or a similar section which is _mapped_ read-only. Attempting to modify such a constant will take an access violation.

To permit this, we should provide a RAII type which temporarily changes the page protection on a given address while in scope, permitting such rewrites to be done safely.

### DIFF
--- a/src/core/protmem.cs
+++ b/src/core/protmem.cs
@@ -19,13 +19,13 @@ public unsafe readonly ref struct FhVirtualProtectScope<T> where T : unmanaged {
     private readonly PAGE_PROTECTION_FLAGS _protection_flags_old;
 
     /// <summary>
-    ///     Changes the memory protection mode beginning at the given <paramref name="address"/> to <paramref name="protection_mode"/>.
+    ///     Changes the memory protection mode beginning at the given address.
     /// </summary>
     /// <remarks>
     ///     The protection of any memory page containing one or more bytes from <paramref name="address"/> to
     ///     (<paramref name="address"/> + sizeof(<typeparamref name="T"/>)) will be changed.
     /// </remarks>
-    /// <param name="address">The starting address to change the memory protection mode at.</param>
+    /// <param name="address">The address to start change the memory protection mode from.</param>
     /// <param name="protection_mode">The memory protection mode to set for the targeted memory region.</param>
     public FhVirtualProtectScope(T* address, PAGE_PROTECTION_FLAGS protection_mode) {
         PAGE_PROTECTION_FLAGS flOldProtect;
--- a/src/core/protmem.cs
+++ b/src/core/protmem.cs
@@ -19,18 +19,24 @@ public unsafe readonly ref struct FhVirtualProtectScope<T> where T : unmanaged {
     private readonly PAGE_PROTECTION_FLAGS _protection_flags_old;
 
     /// <summary>
-    ///     Changes the page protection at the given <paramref name="address"/> to <paramref name="protection_flags"/>.
+    ///     Changes the memory protection mode beginning at the given <paramref name="address"/> to <paramref name="protection_mode"/>.
     /// </summary>
-    public FhVirtualProtectScope(T* address, PAGE_PROTECTION_FLAGS protection_flags) {
+    /// <remarks>
+    ///     The protection of any memory page containing one or more bytes from <paramref name="address"/> to
+    ///     (<paramref name="address"/> + sizeof(<typeparamref name="T"/>)) will be changed.
+    /// </remarks>
+    /// <param name="address">The starting address to change the memory protection mode at.</param>
+    /// <param name="protection_mode">The memory protection mode to set for the targeted memory region.</param>
+    public FhVirtualProtectScope(T* address, PAGE_PROTECTION_FLAGS protection_mode) {
         PAGE_PROTECTION_FLAGS flOldProtect;
-        PInvoke.VirtualProtect(address, (nuint) sizeof(T), protection_flags, &flOldProtect);
+        PInvoke.VirtualProtect(address, (nuint) sizeof(T), protection_mode, &flOldProtect);
 
         _address              = address;
         _protection_flags_old = flOldProtect;
     }
 
     /// <summary>
-    ///     Reverts the page protection at the target address to its previous state.
+    ///     Reverts the memory protection mode of the targeted region to its previous state.
     /// </summary>
     public void Dispose() {
         PAGE_PROTECTION_FLAGS discard;
--- a/src/core/protmem.cs
+++ b/src/core/protmem.cs
@@ -1,0 +1,34 @@
+﻿// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit;
+
+/// <summary>
+///     A disposable structure which changes the protection flags on a memory location
+///     while it remains in scope, reverting to the original flags when disposed.
+/// </summary>
+/// <remarks>
+///     This is useful for rewriting constants embedded into read-only process sections,
+///     which would normally cause an access violation.
+/// </remarks>
+[SupportedOSPlatform("windows5.1.2600")]
+internal unsafe readonly ref struct FhVirtualProtectScope<T> where T : unmanaged {
+    private readonly T*                    _lpAddress;
+    private readonly PAGE_PROTECTION_FLAGS _flOldProtect;
+
+    public FhVirtualProtectScope(T* lpAddress, PAGE_PROTECTION_FLAGS flNewProtect) {
+        PAGE_PROTECTION_FLAGS flOldProtect;
+        PInvoke.VirtualProtect(lpAddress, (nuint) sizeof(T), flNewProtect, &flOldProtect);
+
+        _lpAddress    = lpAddress;
+        _flOldProtect = flOldProtect;
+    }
+
+    public void Dispose() {
+        PAGE_PROTECTION_FLAGS discard;
+        PInvoke.VirtualProtect(_lpAddress, (nuint) sizeof(T), _flOldProtect, &discard);
+    }
+}
+
--- a/src/core/protmem.cs
+++ b/src/core/protmem.cs
@@ -25,7 +25,7 @@ public unsafe readonly ref struct FhVirtualProtectScope<T> where T : unmanaged {
     ///     The protection of any memory page containing one or more bytes from <paramref name="address"/> to
     ///     (<paramref name="address"/> + sizeof(<typeparamref name="T"/>)) will be changed.
     /// </remarks>
-    /// <param name="address">The address to start change the memory protection mode from.</param>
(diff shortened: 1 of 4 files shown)


EvelynTSMG:
Are we sure the compiler never inlined these constants? If it did, while providing this would not be completely pointless, it should come with an extra warning as to its usage in the "remarks" section of the documentation.

fkelava:
`.rdata` is `0xFD800` long. If a constant is ever used more than once, it will be emitted there as X/X-2 HD for Steam were compiled optimized for size. Operating over that space is far from pointless- it ranges from pragmatic to necessary.

ex.
```
006dafb4 dc 0d 80  FMUL  double ptr [DOUBLE_00b4a180]  = 29.96999931335449
         a1 b4 00

DOUBLE_00b4a180  XREF[3]:     UpdateTexture:006d793f(R), 
                              FUN_006d9300:006d933b(R), 
                              updateVideoCallback:006dafb4(R)  
00b4a180 00 00 00        double     29.96999931335449
         e0 51 f8 
         3d 40

// in all three method bodies
uVar3 = +ftoul((uVar3 >> 0x20 & 0x80000000) << 0x20 + ((uVar3 & 0x7fffffffffffffff) * 29.97 * 0.001));
```
Without this, to adjust the operation of these three methods, the only recourse someone has is to reimplement them entirely because the offending statement is within the method body. Rewriting the constant is the sensible approach. If one attempts to using a standard `FhUtil.set_at` as such:
```cs
private static void _f1() {
    double* ptr_rdata_video_target_framerate = FhUtil.ptr_at<double>(FhUtil.select(0x74A180, 0x827620, 0x827620));
    *ptr_rdata_video_target_framerate = 2;
}
```
they take an AV. It is `FhUtil.set_at` and co. that insufficiently warn that you must ensure the target address is actually writable, and they should refer you to this type.
```cs
private static void _f2() {
    double* ptr_rdata_video_target_framerate = FhUtil.ptr_at<double>(FhUtil.select(0x74A180, 0x827620, 0x827620));

    using (FhVirtualProtectScope<double> scope = new(ptr_rdata_video_target_framerate, PAGE_PROTECTION_FLAGS.PAGE_READWRITE)) {
        *ptr_rdata_video_target_framerate *= 2;
    }
}
```

EvelynTSMG (dismissed):

src/core/protmem.cs
```diff
@@ -14,21 +14,33 @@ namespace Fahrenheit;
...
+    private readonly PAGE_PROTECTION_FLAGS _protection_flags_old;
 
-    public FhVirtualProtectScope(T* lpAddress, PAGE_PROTECTION_FLAGS flNewProtect) {
+    /// <summary>
+    ///     Changes the memory protection mode beginning at the given <paramref name="address"/> to <paramref name="protection_mode"/>.
+    /// </summary>
+    /// <remarks>
+    ///     The protection of any memory page containing one or more bytes from <paramref name="address"/> to
+    ///     (<paramref name="address"/> + sizeof(<typeparamref name="T"/>)) will be changed.
+    /// </remarks>
```
EvelynTSMG, line 27: There's a TL;DR at the bottom, and the important bits are in bold.

As per the example for documentation comments in the [Language Spec §D.5.1](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d51-c-source-code), specifically the documentation for the `void Point.Translate(int dx, int dy)` method copied for convenience:
```cs
/// <summary>
/// This method changes the point's location by
/// the given x- and y-offsets.
/// <example>For example:
/// <code>
/// Point p = new Point(3, 5);
/// p.Translate(-1, 3);
/// </code>
/// results in <c>p</c>'s having the value (2, 8).
/// <see cref="Move"/>
/// </example>
/// </summary>
/// <param name="dx">The relative x-offset.</param>
/// <param name="dy">The relative y-offset.</param>
```
The `paramref` tag **should not be used** every time a parameter's name is used – **unless**, as per the example for the tag in the [Language Spec §D.3.10](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/language-specification/documentation-comments#d310-paramref), **the name is specifically used to refer to the value of that parameter**. The tag example is also copied below for convenience:
```cs
/// <summary>
/// This constructor initializes the new Point to
/// (<paramref name="xPosition"/>,<paramref name="yPosition"/>).
/// </summary>
/// <param name="xPosition">the new Point's x-coordinate.</param>
/// <param name="yPosition">the new Point's y-coordinate.</param>
```

(Note: The language spec examples are not necessarily a source of best practices, as exemplified by our not beginning documentation comments with "this method", but it is a useful guide for how one would expect tags to be used and doc comments to be structured.)

### Example and TL;DR
```suggestion
    /// <summary>
    ///     Changes the memory protection mode beginning at the given address.
    /// </summary>
    /// <remarks>
    ///     The protection of any memory page containing one or more bytes from <paramref name="address"/> to
    ///     (<paramref name="address"/> + sizeof(<typeparamref name="T"/>)) will be changed.
    /// </remarks>
```

src/core/protmem.cs
```diff
@@ -14,21 +14,33 @@ namespace Fahrenheit;
...
+    /// <param name="address">The starting address to change the memory protection mode at.</param>
```
EvelynTSMG: This is just a nitpick, but we aren't doing something just *at* the start, we are doing something *from* the start, as per the remarks.
```suggestion
    /// <param name="address">The address to start changing the memory protection mode from.</param>
```

src/core/util.cs
```diff
@@ -21,9 +21,34 @@ internal static T select<T>(T ffx, T ffx2, T ffx2lm, T? defval = default) {
         };
     }
 
-    public static T* ptr_at<T>(nint address)          where T : unmanaged { return (T*)(FhEnvironment.BaseAddr + address); }
-    public static T  get_at<T>(nint address)          where T : unmanaged { return *ptr_at<T>(address);                    }
-    public static T  set_at<T>(nint address, T value) where T : unmanaged { return *ptr_at<T>(address) = value;            }
+    /// <summary>
+    ///     Returns a pointer to <typeparamref name="T"/> at the absolute address obtained by
+    ///     adding the given <paramref name="offset"/> to the base address of the running game's executable.
```
EvelynTSMG: See [my previous review](https://github.com/fahrenheit-crew/fahrenheit/pull/260/changes/2a0cb2bfcc65e738940d2d1571074089f6c9044f..aa7dc924f4ee5b41d24fdc077ba6d81566017440#r3762463595) on usage of the paramref tag.

src/core/util.cs
```diff
@@ -21,9 +21,34 @@ internal static T select<T>(T ffx, T ffx2, T ffx2lm, T? defval = default) {
...
+    /// </summary>
+    /// <param name="offset">The offset, from the running game executable's base address, to return a pointer to <typeparamref name="T"/> at.</param>
+    /// <returns>A pointer to a value of type <typeparamref name="T"/> at the given address.</returns>
+    public static T* ptr_at<T>(nint offset) where T : unmanaged => (T*)(FhEnvironment.BaseAddr + offset);
+
+    /// <summary>
+    ///     Reads a value of type <typeparamref name="T"/> from the address obtained by adding
+    ///     the given <paramref name="offset"/> to the base address of the running game's executable.
```
EvelynTSMG: See [my previous review](https://github.com/fahrenheit-crew/fahrenheit/pull/260/changes/2a0cb2bfcc65e738940d2d1571074089f6c9044f..aa7dc924f4ee5b41d24fdc077ba6d81566017440#r3762463595) on usage of the paramref tag.

src/core/util.cs
```diff
@@ -21,9 +21,34 @@ internal static T select<T>(T ffx, T ffx2, T ffx2lm, T? defval = default) {
...
+    /// <returns>The current value at the given address.</returns>
+    public static T get_at<T>(nint offset) where T : unmanaged => *ptr_at<T>(offset);
+
+    /// <summary>
+    ///     Writes a given <paramref name="value"/> of type <typeparamref name="T"/> to the address obtained
+    ///     by adding the given <paramref name="offset"/> to the base address of the running game's executable.
+    /// </summary>
+    /// <remarks>
+    ///     The target <paramref name="offset"/> must be in a writable memory region.
+    ///     If it is not, you must use a <see cref="FhVirtualProtectScope{T}"/>.
```
EvelynTSMG: In this case, I believe it is worth mentioning the reason *why* the user **must** use `FhVirtualProtectScope<T>`:
```suggestion
    ///     If it is not, you must use a <see cref="FhVirtualProtectScope{T}"/> to avoid an <c>AccessViolationException</c>.
```
\
I am not sure whether it would be best to use `<c>` or `<see>`.

src/core/util.cs
```diff
@@ -21,9 +21,34 @@ internal static T select<T>(T ffx, T ffx2, T ffx2lm, T? defval = default) {
...
+    /// </remarks>
+    /// <param name="offset">The offset, from the running game executable's base address, to write the given <paramref name="value"/> to.</param>
```
EvelynTSMG: See [my previous review](https://github.com/fahrenheit-crew/fahrenheit/pull/260/changes/2a0cb2bfcc65e738940d2d1571074089f6c9044f..aa7dc924f4ee5b41d24fdc077ba6d81566017440#r3762463595) on usage of the paramref tag.

Although clunky, I believe `to write <paramref name="value"/> to` would work here as well. I would, however, prefer simply removing the `<paramref>`.

src/core/util.cs
```diff
@@ -21,9 +21,34 @@ internal static T select<T>(T ffx, T ffx2, T ffx2lm, T? defval = default) {
...
+    /// <param name="value">The value of type <typeparamref name="T"/> to write at the given address.</param>
+    /// <returns>The previous value at the given address.</returns>
```
EvelynTSMG, line 50: I'm pretty sure this just returns the value it was set to, actually. Whoops.
I believe the assignment operator returns the new value, no?

EvelynTSMG (changes requested):

EvelynTSMG:
Overall, amazing work! I have a few nitpicks but that's all they are, so don't stress yourself out about them.

src/core/util.cs
fkelava: `<see>` lets one see this when hovering it:
> The exception that is thrown when there is an attempt to read or write protected memory.

which I think is more useful.

src/core/protmem.cs
```diff
@@ -19,13 +19,13 @@ namespace Fahrenheit;
...
     /// <summary>
-    ///     Changes the memory protection mode beginning at the given <paramref name="address"/> to <paramref name="protection_mode"/>.
+    ///     Changes the memory protection mode beginning at the given address.
     /// </summary>
     /// <remarks>
     ///     The protection of any memory page containing one or more bytes from <paramref name="address"/> to
     ///     (<paramref name="address"/> + sizeof(<typeparamref name="T"/>)) will be changed.
     /// </remarks>
-    /// <param name="address">The starting address to change the memory protection mode at.</param>
+    /// <param name="address">The address to start change the memory protection mode from.</param>
```
EvelynTSMG: "Start change" is two verbs in a row, it should be "changing"

EvelynTSMG (changes requested):

EvelynTSMG (approved):
Great work! And hopefully a good learning experience long-term as well 😊

PR #263, opened 2026-08-14
Upstream delegates for Warp

fkelava:
Before [Warp](https://github.com/fkelava/warp) can be sent off to live in its own repository, it requires the calls it intends to use to be upstreamed. This PR tracks that work, parts of which are still pending.


src/core/ffx/actor.cs
```diff
@@ -5,57 +5,63 @@
...
+    [FieldOffset(0x634)] public Vector4    chr_optpos_vec_16;
+    [FieldOffset(0x644)] public Vector4    chr_optpos_vec_17;
+    [FieldOffset(0x654)] public Vector4    chr_optpos_vec_18;
+    [FieldOffset(0x664)] public Vector4    chr_optpos_vec_19;
+    [FieldOffset(0x674)] public Vector4    chr_optpos_vec_20;
+    [FieldOffset(0x684)] public Vector4    chr_optpos_vec_21;
+    [FieldOffset(0x694)] public Vector4    chr_optpos_vec_22;
+    [FieldOffset(0x6A4)] public Vector4    chr_optpos_vec_23;
+    [FieldOffset(0x6B4)] public Vector4    chr_optpos_vec_24;
+    [FieldOffset(0x6C4)] public Vector4    chr_optpos_vec_25;
```
DawnVeilwinter: Could we make this an InlineArray please?

src/core/ffx/call.cs
```diff
@@ -21,6 +21,191 @@ namespace Fahrenheit.FFX;
...
+        => new( new FhMethodLocation("FFX.exe", 0x679510) );
+
+    public static FhMethodHandle<Fahrenheit.FhCall.d_CT_RetInt> CT_504C_RetInt
+        => new( new FhMethodLocation("FFX.exe", 0x6786A0) );
+
+    public static FhMethodHandle<Fahrenheit.FhCall.d_CT_RetInt> CT_0001_RetInt
+        => new( new FhMethodLocation("FFX.exe", 0x45CE70) );
+
+    public static FhMethodHandle<Fahrenheit.FhCall.d_CT_Init> CT_5010_Init
+        => new( new FhMethodLocation("FFX.exe", 0x679820) );
```
DawnVeilwinter, line 31: These should be sorted:
- `CT_ABCD` before `CT_EFGH` when `ABCD < EFGH`
- `Init` before `Exec` before `RetFloat` before `RetInt`

src/core/ffx2/call.cs
```diff
@@ -22,4 +22,14 @@ public static unsafe partial class FhCall {
     internal static FhMethodHandle<d_GetLastMissionJobName> GetLastMissionJobName
         => new( new FhMethodLocation("FFX-2.exe", 0x368570) );
 
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public delegate void d_Ch_SetMotionSpeed(uint ptr_actor, ushort speed);
```
DawnVeilwinter, line 26: I assume you don't have a defined type for the `ptr_actor` here? Otherwise, FhCall delegates are never called outside of the games running, so this can be `Actor*` or whatever other type.

src/core/ffx/call.cs
```diff
@@ -21,6 +21,191 @@ namespace Fahrenheit.FFX;
...
+    public static FhMethodHandle<d_MsEffectSetSpeed> MsEffectSetSpeed
+        => new( new FhMethodLocation("FFX.exe", 0x3884F0) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public delegate void d_graphicDrawMainMenuWaterEffect();
+    public static FhMethodHandle<d_graphicDrawMainMenuWaterEffect> graphicDrawMainMenuWaterEffect
+        => new( new FhMethodLocation("FFX.exe", 0x23EAD0) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public delegate void d_Sg_EffectRoutineEnable(sbyte arg1);
```
DawnVeilwinter: This is just a `byte`, there's nothing to indicate it's signed.

src/core/ffx/call.cs
```diff
@@ -21,6 +21,191 @@ namespace Fahrenheit.FFX;
...
+    public static FhMethodHandle<d_AtelGetCharHandle> AtelGetCharHandle
+        => new( new FhMethodLocation("FFX.exe", 0x46AE10) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public delegate void d_Ch_CalcAnim(float delta);
+    public static FhMethodHandle<d_Ch_CalcAnim> Ch_CalcMain
+        => new( new FhMethodLocation("FFX.exe", 0x432E90) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public unsafe delegate void d_Ch_SetMotionSpeed(Actor* ptr_actor, ushort speed);
```
DawnVeilwinter, line 200: Usually we'd just type `Actor* actor`, since it's not dereferenced in the function.

src/core/ffx/call.cs
```diff
@@ -21,6 +21,191 @@ namespace Fahrenheit.FFX;
...
+    public static FhMethodHandle<d_MsSetChrStatInfo> MsSetChrStatInfo
+        => new( new FhMethodLocation("FFX.exe", 0x3B4B80) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public delegate void d_FUN_003B4AA0(uint chr_id, uint stat_id, float value);
+    public static FhMethodHandle<d_FUN_003B4AA0> FUN_003B4AA0
+        => new( new FhMethodLocation("FFX.exe", 0x3B4AA0) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public unsafe delegate void d_Ch_SetKeepFps(Actor* ptr_actor, uint keep_fps);
```
DawnVeilwinter, line 150: Usually we'd just type `Actor* actor`, since it's not dereferenced in the function.

DawnVeilwinter (changes requested):

src/core/ffx2/call.cs
fkelava: It's not yet defined.

DawnVeilwinter (approved):
Thanks

PR #266, opened 2026-08-21
Stage1: Change working directory preemptively

fkelava:
There is a strange, unspoken detail about Fahrenheit initialization: `init` runs with the working directory still set to `fahrenheit/bin`, and is then changed immediately thereafter. This means that relative paths can resolve to two different places before and after `init`, which can range from being merely unexpected to having serious consequences.

The other consequence thereof is that anything relying on the game's working directory being 'normally' set, most notably [ffgriever's EFL](https://gitlab.com/ffgriever/ffx-x-2-hd-external-file-loader), can't work. But it doesn't have to be like this.

We _can_, in fact, fix this; `fhstage1` is guaranteed to load first, and we can instantly change the working directory instead of doing it after initializing Fahrenheit. The rest of Stage1 and Fahrenheit must be amended accordingly.

I also took the liberty of splitting out S1 EH into a separate code file for ease of maintenance.

### DIFF
--- a/src/stage1/src/main.cpp
+++ b/src/stage1/src/main.cpp
@@ -9,59 +9,51 @@
  * See THIRD-PARTY-NOTICES.
  *
  * For the HostFXR bits, see https://github.com/dotnet/samples/blob/main/core/hosting/src/NativeHost/nativehost.cpp.
- * For the exception handling, see:
- * - http://code.aaronballman.com/minidumper/MiniDump.cpp
- * - https://github.com/folgerwang/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Private/Windows/WindowsPlatformCrashContext.cpp
  */
 
 #include "fhstage1.h"
 
-// Function pointer to managed delegate with our own signature
-typedef void (CORECLR_DELEGATE_CALLTYPE* fh_init)();
+typedef void (CORECLR_DELEGATE_CALLTYPE* fh_init)(); // Function pointer to managed delegate with our own signature
 
-using string_t = std::basic_string<char_t>;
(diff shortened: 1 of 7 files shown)


src/stage1/src/main.cpp
```diff
@@ -9,227 +9,61 @@
...
+char_t path_fh_buf    [MAX_PATH]; // The path to the `fahrenheit/bin` directory we were started in.
 
-// Globals to hold hostfxr exports
 hostfxr_initialize_for_runtime_config_fn g_fnptr_hostfxr_init;
 hostfxr_set_runtime_property_value_fn    g_fnptr_hostfxr_set_runtime_property;
 hostfxr_get_runtime_delegate_fn          g_fnptr_hostfxr_get_delegate;
 hostfxr_close_fn                         g_fnptr_hostfxr_close;
 
-// Using the nethost library, discover the location of hostfxr and get exports
+BOOL stage1_eh_install(LPBYTE ptr_main_module); // Forward declaration of EH installer function
```
DawnVeilwinter, line 31: This comment should probably be on a separate line before this declaration and just read "Forward declarations"

DawnVeilwinter (dismissed):
One minor formatting nitpick, feel free to just dismiss the review and I'll approve if that is your will

src/stage1/src/main.cpp
fkelava: I agree that it should, but VS is silly and attaches a comment to the nearest declaration it can find. That's why I ended up making separate comments for each global over in `main.cpp`, and that's why this is as it is.

DawnVeilwinter (approved):

PR #267, opened 2026-08-21
EFL: Allow "multi-game" operation and cross-loading of the inactive game's assets

fkelava:
Fahrenheit uniquely permits "multi-game" mods, but the EFL has only ever operated on the currently active game, on the tacit and thus far undisputed assumption that only the active game's assets are available.

But in reality, we are not beholden to such limits. We've reversed enough of the game's file loading layer to permit access to _both games' assets_ from either game. In other words, when playing FF X, one could utilize FF X-2 assets in Fahrenheit and vice versa. From this it also follows the EFL must be able to effect replacements to both games' assets in either game.

This PR enables both. As a bonus, the game's unwieldy `../../../` stream prefix is trimmed to simply `/`, to allow for easier and more legible addressing for end users.


DawnVeilwinter (approved):
I assume the delegates are correct.
Documentation looks good.
Code was previously demonstrated to function over on Discord.

Good work!

PR #268, opened 2026-08-23
Fix the game's defective path normalization

fkelava:
In #267 we permitted cross-loading of the inactive game's assets. This unraveled due to twofold stupidity in the remaster:
- Some paths are not normalized at all.
- Some paths have a hard-coded old stream prefix and the path normalizer does not correct it.

We must therefore employ the nuclear option and fix both. The game's path normalizer is reduced to a copying stub, and combined normalization (EFL + game default) is performed at the point of file opening.


DawnVeilwinter (approved):
Looks good, feel free to merge if/when tested

PR #275, opened 2026-08-25
Simplify the 'allocator fix' module somewhat

fkelava:
Move out the copious amount of debug prints into a little ImGui display tracking allocation stats. If indeed one needs per-allocation tracking, it can be added and kept locally as needed.

Additionally, consolidate things into a single hook, and fix an issue where reverse reservations were not tracked in the global total.

### DIFF
--- a/src/core/native/NativeMethods.txt
+++ b/src/core/native/NativeMethods.txt
@@ -23,6 +23,7 @@ HMODULE
 // Memory protection modifier
 
 VirtualProtect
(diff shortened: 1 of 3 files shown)


DawnVeilwinter (approved):
Looks good!

The UI is acceptable for now, though should eventually be merged into debug mod instead I think. Unconditionally opening a window when compiled for debug isn't great.

PR #276, opened 2026-08-26
Fix yet another case of defective path normalization

fkelava:
In yet another installment of the rabbit hole without ending, it turns out that path handling in these games (and particularly X-2/LM) is held together by duct tape and dreams. The only reason it ever worked by default was the hardcoded stream prefix.

Here we fix two additional cases where the developers decided path normalization was optional when it really isn't.


src/runtime/fileloader.cs
```diff
@@ -240,26 +241,50 @@ private void h_sf_sp_set(byte* ptr_stream_prefix) {
...
-     * - /FFX_Data/GameData/PS3Data/chr/mon/m220/fp/tex/GCM/16128_0_0_8_256_128.dds.phyre
-     * - /ffx_ps2/ffx/master/new_depc
-     * - /help/test_proj/test_proj_page.sps2
-     * - ../../../ffx_ps2/ffx/proj/map/masaki/
+     * Buckle up. This is where things get bad.
+     * 
+     * Simplifying the stream prefix, on paper, should not be problematic. The game has a path normalizer
+     * function `fiosUnifyFilename`, so at worst we have to reimplement just that, right?
+     * 
+     * The game's actual _use_ of path normalization can most charitably be described as completely batshit insane.
```
DawnVeilwinter: I'm surprised you managed to keep the cursing back on Discord and in the actual PR, but not here. lol
```suggestion
     * The game's usage of path normalization can, charitably, be described as inconsistent at best.
```

DawnVeilwinter (changes requested):

DawnVeilwinter (approved):
Looks good! Another rabbit hole patched up :3

PR #277, opened 2026-08-26
X-2 Save Header: Define `times_played` and `level_job_lm`

cxldalyy:
`level_job_lm` quite simply is the level of the character's currently equipped dressphere/job in Last Mission. It also uses a "Lv." prefix, similarly to `level_chr_lm`.
`times_played` obviously referring to the number of playthroughs a player has completed. For Last Mission, this value is always 0, since it's not considered a "main game" playthrough. For X-2, this value is always at least 1, and increments whenever starting a New Game Plus.


DawnVeilwinter (approved):

fkelava (approved):
Thanks!

PR #278, opened 2026-08-27
Short-circuit subsequent texture (un)loads

fkelava:
Due to some gratuitous stupidity on my behalf, `load_*_texture` would return `true` on a successful load but then `false` on a subsequent call, even though the texture was already loaded. This makes it confusing to impossible to write a simple load-all routine, so let's do that correctly now.


DawnVeilwinter (approved):
I do think it could be simplified, but very well.
Good work and thank you for inadvertently fixing a bug I hadn't noticed!

PR #279, opened 2026-08-28
Minor buildsystem improvements

fkelava:
- Stop emitting `vcpkg` artifacts under `src/stage{0|1}/vcpkg_installed`. They belong with other `artifacts`.
- Update the `.gitignore` template, and amend some outdated entries.
- Add a little README hinting that `env` exists, and what it is for.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 *.user
 *.userosscache
 *.sln.docstates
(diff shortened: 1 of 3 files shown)


DawnVeilwinter (approved):
Seems good, thanks

PR #283, opened 2026-08-31
Add an improved soft-reset function

fkelava:
For the small price of three common functions, we are able to construct an improved soft-reset that works in most permutations of field, combat and FMV. 

Down the road, we may wish to separate this out into its own mod, or an EFP thing. For now, ship it in the runtime so users can access it readily.


src/core/call.cs
```diff
@@ -34,6 +34,21 @@ public static unsafe partial class FhCall {
      * circumvent it. We simply refuse to support any such scenario.
      */
 
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public unsafe delegate void d_AtelJumpGameOver();
+    public static FhMethodHandle<d_AtelJumpGameOver> AtelJumpGameOver 
+        => new( new FhMethodLocation(0x46D9A0, 0x3283A0) );
+
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public unsafe delegate uint d_MsBattleCheck();
```
EvelynTSMG, line 43: If a function returns only 0 or 1 as integers, it may be returning a `BOOL`, which I believe to be the case here. `BOOL` is defined by Windows headers as `int`; therefore, this should return `int`, not `uint`.

src/runtime/reset.cs
```diff
@@ -0,0 +1,39 @@
...
+    public override void render_imgui() {
+        bool activated = (ImGui.IsKeyDown(ImGuiKey.GamepadL1) && ImGui.IsKeyDown(ImGuiKey.GamepadR1) && ImGui.IsKeyPressed(ImGuiKey.GamepadStart))
+                      || (ImGui.IsKeyDown(ImGuiKey.R)         && ImGui.IsKeyDown(ImGuiKey.S)         && ImGui.IsKeyPressed(ImGuiKey.T));
+
+        if (!activated)
+            return;
+
+        if (FhCall.MsBattleCheck.fnptr!() != 0) {
+            FhUtil.set_at(FhGlobal.game_id is FhGameId.FFX ? 0xD2C9F1 : 0x9F94B5, 1U);
+            FhUtil.set_at(FhGlobal.game_id is FhGameId.FFX ? 0xD2A8E0 : 0x9F78A0, 2U);
```
EvelynTSMG: If you know, please add comments explaining what this is doing.

EvelynTSMG: Also, I'd expect these to use `FhUtil.select`, which should be amended to have the LM parameter be optional? `T? ffx2_lm = null`, which is then immediately set to the ffx2 value if missing with `ffx2_lm ??= ffx2`. That way you have no incentive to avoid using it like this.

EvelynTSMG (changes requested):

EvelynTSMG (approved):
Good work!

DawnVeilwinter (approved):
Good work! From the right account this time!

PR #291, opened 2026-09-03
Equalize capitalization in crossload scenarios

fkelava:
As it turns out, VBF loading is very much case sensitive; it expects things to be all-lowercase. In crossloading (and only in crossloading), I neglected to account for that.


DawnVeilwinter (approved):
I'm assuming that function call lowercases the path? Good good

PR #292, opened 2026-09-03
Fix up a bunch of docstrings, and some formatting

fkelava:


src/core/bootloader.cs
```diff
@@ -154,12 +154,11 @@ internal FhLoader() {
         FhInternal.Log.LogDirect($"----");
     }
 
-    /// <summary>
-    ///     Attempts to map a <see cref="AssemblyName"/> to an already loaded Fahrenheit DLL <see cref="Assembly"/>.
-    ///     <para/>
+    /// <summary>Attempts to map a <see cref="AssemblyName"/> to an already loaded Fahrenheit DLL <see cref="Assembly"/>.</summary>
+    /// <remarks>
     ///     This is because Fahrenheit mod DLLs are not permitted to bundle other mod DLLs they depend on;
     ///     whichever version of the dependency the user actually has installed will be loaded instead.
```
DawnVeilwinter: This is a bit confusing as-is, I'm not actually sure what you mean. I know the effect is to not load mod dependencies that are other mods and instead use the already loaded ones, but "the user actually has installed" doesn't parse for me. Maybe "we instead connect it to the already-loaded dependency", or something like that?

src/core/encoding.cs
```diff
@@ -228,29 +263,37 @@ private static bool _extend_cjk(ref Rune code_point) {
...
+    ///     Determines whether the given <paramref name="lang"/> is a CJK language.
     /// </summary>
-    private static bool _is_cjk(FhLangId language) {
-        return language is FhLangId.Chinese
+    private static bool _is_cjk(
+        FhLangId lang
+    ) {
+        return lang is FhLangId.Chinese
                         or FhLangId.Japanese
                         or FhLangId.Korean;
```
DawnVeilwinter: Surely we align the `FhLangId` here?

src/core/encoding.cs
```diff
@@ -468,10 +515,14 @@ 0x100 or
     /// <summary>
     ///     Given some DEdit syntax text in <paramref name="src"/>, computes the size of the buffer required to write all indices
```
DawnVeilwinter: I believe we shouldn't directly mention tools in Core unless necessary; here preferring simply "text", with maybe a `<seealso>` tag linking to a larger explanation of the special syntax.

src/core/encoding.cs
```diff
@@ -491,11 +542,16 @@ public static int compute_index_buffer_size(in ReadOnlySpan<byte> src, FhTextInd
...
     ///     Given some game-encoded text in <paramref name="src"/>, writes to <paramref name="dest"/> the indices
     ///     of a given <paramref name="index_type"/> necessary for the game to access that text.
-    ///     <para/>
-    ///     <paramref name="dest"/> must be properly sized. To obtain the correct size, perform
-    ///     <see cref="compute_index_buffer_size(ReadOnlySpan{byte}, FhTextIndexType)"/> on the source DEdit text.
     /// </summary>
-    public static void write_indices(in ReadOnlySpan<byte> src, Span<byte> dest, FhTextIndexType index_type) {
+    /// <remarks>
+    ///     <paramref name="dest"/> must be properly sized. To obtain the correct size, perform
+    ///     <see cref="compute_index_buffer_size(in ReadOnlySpan{byte}, FhTextIndexType)"/> on the source DEdit text.
```
DawnVeilwinter: I believe we shouldn't directly mention tools in Core unless necessary, particularly when the "source DEdit text" isn't even sourced *from* DEdit; here preferring simply "text", with maybe a `<seealso>` tag linking to a larger explanation of the special syntax.

src/core/excel.cs
```diff
@@ -28,10 +28,11 @@ public struct ExcelSimplifiableTextOffset {
...
     /// <summary>
-    ///     The offset to the simplified text. In Japanese, this would
-    ///     have been hiragana; in Western encodings, it has no effect.
-    ///     <para/>
-    ///     This is completely unused.
+    ///     The offset to the simplified text.
     /// </summary>
+    /// <remarks>
+    ///     In Japanese, this would have been hiragana; 
+    ///     in Western encodings, it has no effect. This is completely unused.
```
DawnVeilwinter: For the sake of reading non-rendered documentation, I think the "This is completely unused." sentence should be on a separate line.

src/core/ffx/excel.cs
```diff
@@ -13,11 +13,8 @@ namespace Fahrenheit.FFX;
  * from player command definitions to aeon stat growth curves.
  */
 
-/// <summary>
-///     The prologue of an Excel container. Describes it in enough detail to construct a reader.
-///     <para/>
-///     To iterate over its contents, use an <see cref="ExcelReader{T}"/>.
-/// </summary>
+/// <summary>The prologue of an Excel container. Describes it in enough detail to construct a reader.</summary>
+/// <remarks>To iterate over its contents, use an <see cref="ExcelReader{T}"/>.</remarks>
```
DawnVeilwinter, line 17: Optionally, I believe a `<seealso cref="ExcelReader{T}"/>` would be perfectly sufficient.

src/core/ffx2/excel.cs
```diff
@@ -13,11 +13,8 @@ namespace Fahrenheit.FFX2;
  * from player command definitions to aeon stat growth curves.
  */
 
-/// <summary>
-///     The prologue of an Excel container. Describes it in enough detail to construct a reader.
-///     <para/>
-///     To iterate over its contents, use an <see cref="ExcelReader{T}"/>.
-/// </summary>
+/// <summary>The prologue of an Excel container. Describes it in enough detail to construct a reader.</summary>
+/// <remarks>To iterate over its contents, use an <see cref="ExcelReader{T}"/>.</remarks>
```
DawnVeilwinter, line 17: Optionally, I believe a `<seealso cref="ExcelReader{T}"/>` would be perfectly sufficient.

src/core/handles.cs
```diff
@@ -109,9 +108,8 @@ unsafe public FhMethodLocation(void* abs_addr) {
 
     /// <summary>
     ///     Gets the address of the module with the given <paramref name="module_name"/>.
-    ///     <para/>
-    ///     If the module is not loaded, the return value is zero.
     /// </summary>
+    /// <remarks> If the module is not loaded, the return value is zero.</remarks>
```
DawnVeilwinter:
```suggestion
    /// <returns>
    ///      The address of the specified module.
    ///      If the module is not loaded, returns <c>0</c>.
    /// </returns>
```

("zero" as a word would also be fine if using the original wording of "the return __value__ is", I just prefer to write it as a value, and prefer this wording.)

src/core/handles.cs
```diff
@@ -121,9 +119,8 @@ private static nint get_module_addr(string module_name) {
     /// <summary>
     ///     Gets the address of a named <paramref name="export"/>
     ///     in the module at address <paramref name="module_addr"/>.
-    ///     <para/>
-    ///     If it does not exist, the return value is zero.
     /// </summary>
+    /// <remarks>If it does not exist, the return value is zero.</remarks>
     private static bool get_export(nint module_addr, string export, out nint ptr_fn) {
```
DawnVeilwinter, line 125: This documentation should account for the `out nint ptr_fn` parameter and stop calling it the return value as if the `bool` didn't exist. This may involve providing individual documentation for each parameter through `<param>`.

src/core/petypes.cs
```diff
@@ -186,11 +186,10 @@ internal unsafe struct PRedBlackTreeNode<T> where T : unmanaged {
  * RTTI inheritance graph -> +79265C
  */
 
-/// <summary>
-///     Describes a unique type in the Phyre type system- its name, size, alignment, and more.
-///     <para/>
+/// <summary>Describes a unique type in the Phyre type system- its name, size, alignment, and more.</summary>
+/// <remarks>
 ///     Classes have a derived <see cref="PClassDescriptor"/> instead, providing information about their members, layout and inheritance chain.
```
DawnVeilwinter: You're missing an oxford comma here.

src/core/save_ui.cs
DawnVeilwinter, line 1: Why the rename? This is specifically a renderer for the UI; `save_ui` might imply it contains types pertaining to the original save display. Maybe we should fully type out `save_ui_renderer`?

src/core/state.cs
```diff
@@ -18,12 +18,11 @@ namespace Fahrenheit;
  * Modules also have access to an analogous 'global state' file, which is not unique per save game.
  */
 
-/// <summary>
-///     Implements the 'local state' mechanism of Fahrenheit.
-///     <para/>
+/// <summary>Implements the 'local state' mechanism of Fahrenheit.</summary>
+/// <remarks>
 ///     In your module, implement <see cref="FhModule.load_local_state(FileStream, FhLocalStateInfo)"/>
 ///     and <see cref="FhModule.save_local_state(FileStream)"/>.
```
DawnVeilwinter: I think `save_local_state` should be mentioned first since without it `load_local_state` has nothing to load.

src/runtime/resources.cs
```diff
@@ -5,11 +5,8 @@
 
 namespace Fahrenheit.Runtime;
 
-/// <summary>
-///     Loads textures and other resources at runtime.
-///     <para/>
-///     In your module, call <see cref="FhApi.Resources"/>.
-/// </summary>
+/// <summary>Loads textures and other resources at runtime.</summary>
+/// <remarks>In your module, call <see cref="FhApi.Resources"/>.</remarks>
```
DawnVeilwinter: I think better wording is possible:
```suggestion
/// <remarks>In your module, you can use this functionality through <see cref="FhApi.Resources"/>.</remarks>
```

src/runtime/imgui.cs
```diff
@@ -14,11 +14,8 @@ namespace Fahrenheit.Runtime;
  * to the draw list for the frame, and those additions are rendered over the game.
  */
 
-/// <summary>
-///     Provides the ability to use the ImGui GUI toolkit within the game.
-///     <para/>
-///     In your module, implement <see cref="FhModule.render_imgui"/>.
-/// </summary>
+/// <summary>Provides the ability to use the ImGui GUI toolkit within the game.</summary>
+/// <remarks>In your module, implement <see cref="FhModule.render_imgui"/>.</remarks>
```
DawnVeilwinter: I think better wording is possible:
```suggestion
/// <remarks>In your module, you can implement <see cref="FhModule.render_imgui"/> to use this functionality.</remarks>
```

DawnVeilwinter (changes requested):
Just some minor changes I'd like to see here and there!

DawnVeilwinter (dismissed):
Looks good! Thank you for the housekeeping ^.^

src/core/save_mgr.cs
```diff
@@ -74,7 +74,7 @@ public sealed class FhSaves {
     private readonly string                  _sm_path_base;
     private readonly string                  _sm_path_default_set;
     private readonly List<FhSaveDisplayData> _sm_display_data;
-    private readonly Renderers               _sm_renderers;
+    private readonly UserInterfaces          _sm_user_interfaces;
```
DawnVeilwinter: I feel like `_sm_uis` would have sufficed? If you disagree, resolve this without changes and I'll approve.

DawnVeilwinter (changes requested):

DawnVeilwinter (approved):

PR #296, opened 2026-09-05
Fix `sphere_effect` type from `byte` to `ushort` enum `SphereTargets`

### LINKED ISSUE #295: SphereGridNodeType's sphere effect is the wrong size
We currently provide `SphereGridNodeType.sphere_effect` as a byte:
```cs
[StructLayout(LayoutKind.Sequential)]
public struct SphereGridNodeType {
    public ExcelSimplifiableTextOffset name;
    public ExcelSimplifiableTextOffset help;
    public byte                        sphere_effect;
    public T_XCommandId                ability_id;
    public ushort                      amount;
    public ushort                      icon_id;
}
```
However, as noted by Rurusachi over on Discord, it is in fact a 2-byte value:

Additionally, it appears to be a bitfield, so we should provide it as an enum. In fact, this seems to be the same as the `SphereTargets` enum we already define elsewhere.

cxldalyy:
Closes #295

### DIFF
--- a/src/core/ffx/panel.cs
+++ b/src/core/ffx/panel.cs
@@ -9,7 +9,7 @@ namespace Fahrenheit.FFX;
 public struct SphereGridNodeType {
     public ExcelSimplifiableTextOffset name;
     public ExcelSimplifiableTextOffset help;
-    public byte                        sphere_effect;
+    public SphereTargets               sphere_effect;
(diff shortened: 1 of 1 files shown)


fkelava (approved):
Thanks.

DawnVeilwinter (approved):
Looks good to me, thank you!

PR #297, opened 2026-09-08
Fix errors identified in `a11` testing

fkelava:
Per AP testing.

### DIFF
--- a/src/runtime/malloc.cs
+++ b/src/runtime/malloc.cs
@@ -36,8 +36,8 @@ namespace Fahrenheit.Runtime;
 [SupportedOSPlatform("windows5.1.2600")]
 public unsafe sealed class FhMallocModule : FhModule {
 
-    private static nuint _reserved { 
-        get => FhUtil.get_at<nuint>(FhUtil.select(0x153CD44, 0x14E6AB4, 0x14E6AB4)); 
+    private static nuint _reserved {
+        get => FhUtil.get_at<nuint>(FhUtil.select(0x153CD44, 0x14E6AB4, 0x14E6AB4));
         set => FhUtil.set_at       (FhUtil.select(0x153CD44, 0x14E6AB4, 0x14E6AB4), value);
     }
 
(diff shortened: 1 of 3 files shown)


DawnVeilwinter (commented):
We should await further playtesting to confirm long-term efficacy of the malloc fix. That is to say, I will approve this only after tomorrow's playtest, which will be built from this PR's branch.

src/runtime/malloc.cs
```diff
@@ -99,30 +99,22 @@ private void h_mpool_init() {
...
-     * reserved pool still congests the low reaches of address space. Can we do better?
-     *
-     * The game eventually defers any action by its primary allocator to VirtualAlloc/VirtualFree.
-     * See https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc,
-     * https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualfree.
-     *
-     * By passing MEM_TOP_DOWN, VirtualAlloc can be induced to go top-down; that is, start allocating
-     * from {7|F}FFF_FFFF downwards. This applies to the primary pool as well!
-     * When the game is 4G patched, this leaves a much larger free block under 2G.
+     * This bug recurs regardless as long as the 4GB patch is applied, just less frequently.
```
DawnVeilwinter: This bug does not occur at all in known history, and is only theoretically possible if we don't force it with top-down allocation.

src/runtime/malloc.cs
```diff
@@ -99,30 +99,22 @@ private void h_mpool_init() {
...
+     * Therefore we _temporarily_ disable top-down allocation.
```
DawnVeilwinter:
```suggestion
     * Therefore, we must give up on top-down allocation until all plausibly affected pointers
       can safely exist in the address space above 0x7FFF_FFFF.
```

DawnVeilwinter (changes requested):
Just some wording nitpicks

DawnVeilwinter (approved):

PR #301, opened 2026-09-14
Add sound effects to the save UI

### LINKED ISSUE #300: The save UI is missing sound effects
Sound effects can be played using the `SndSepPlaySimple(uint sound_id)` or `SndSepPlay(uint sound_id, uint pan, uint volume)` functions. Calling `SndSepPlaySimple(uint sound_id)` is equivalent to calling `SndSepPlay(uint sound_id, 63, 127)`.

DawnVeilwinter:
Fixes #300.


src/core/ffx/call.cs
```diff
@@ -1206,10 +1206,15 @@ public static FhMethodHandle<d_AtelGetEventName> AtelGetEventName
         => new( new FhMethodLocation("FFX.exe", 0x4796E0) );
 
     [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
-    public delegate void d_SndSepPlaySimple(uint arg1);
+    public delegate void d_SndSepPlaySimple(uint sound_id);
     public static FhMethodHandle<d_SndSepPlaySimple> SndSepPlaySimple
         => new( new FhMethodLocation("FFX.exe", 0x486DE0) );
```
fkelava: This function is common, returns `undefined4`, and should be fused with its X-2 definition. `SndPlay`/`SqsSePlay` do not return `void`, but `undefined4`. The compiler took liberties with the calling convention.

src/core/ffx/call.cs
```diff
@@ -1206,10 +1206,15 @@ public static FhMethodHandle<d_AtelGetEventName> AtelGetEventName
...
 
+    [UnmanagedFunctionPointer(CallingConvention.Cdecl)]
+    public delegate void d_SndSepPlay(uint sound_id, uint volume, uint pan);
+    public static FhMethodHandle<d_SndSepPlay> SndSepPlay
+        => new( new FhMethodLocation("FFX.exe", 0x486DB0) );
+
```
fkelava: The same as above. This does not return `void` but `undefined4` and is common with X-2. Fuse it.

fkelava (changes requested):

src/core/ffx/ids/sound.cs
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
```
fkelava, line 1: `src/core/ffx/ids/sound.cs` seems like a fine name for this file. Analogously X-2.

src/core/ffx/ids/snd_id.cs
```diff
@@ -0,0 +1,31 @@
...
+    public const uint PAGE_TURN   = 0x80000030;
+
+    public const uint BTL_TURN_START = 0x80000002;
+    public const uint BTL_SHATTER    = 0x80000031;
+    public const uint BTL_FADE_IN    = 0x80000065;
+
+    public const uint LEVEL_UP_CHIME = 0x80000017;
+
+    public const uint CHEST_OPEN    = 0x80000009;
+    public const uint TRESURE_CHIME = 0x80000026;
```
fkelava:
```suggestion
    public const uint TREASURE_CHIME = 0x80000026;
```

src/core/ffx/ids/sound.cs
```diff
@@ -0,0 +1,31 @@
...
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit.FFX.Ids;
+
+public class SoundId {
+    public const uint UI_ACTION   = 0x80000001;
```
fkelava, line 9: Consider cutting these up as we do for some other IDs: `0x8000_0001` and such. Optional.

src/core/audio.cs
```diff
@@ -0,0 +1,34 @@
...
+    ///     Defaults to <c>0.5</c>, playing from both the left and right equally.
+    /// </param>
+    public int play_sound(uint sound_id, float volume = 1.0f, float pan = 0.5f) {
+        //TODO: Ensure the sound bank for the sound is loaded
+
+        volume = float.Clamp(volume, 0.0f, 1.0f);
+        pan    = float.Clamp(pan,    0.0f, 1.0f);
+
+        uint int_volume = uint.CreateSaturating(volume * 127);
+        uint int_pan    = uint.CreateSaturating(pan * 127);
```
fkelava:
```suggestion
        uint int_volume = uint.CreateSaturating(volume * 127);
        uint int_pan    = uint.CreateSaturating(pan    * 127);
```

src/core/audio.cs
```diff
@@ -0,0 +1,34 @@
+// SPDX-License-Identifier: LGPL-3.0-or-later
+//
+// This file is part of Fahrenheit, © 2023-2026 The Fahrenheit contributors.
+// It is licensed to you under the GNU Lesser General Public License, version 3.0 or later. See COPYING, COPYING.LESSER.
+
+namespace Fahrenheit;
+
+public class FhAudio {
```
fkelava:
```suggestion
public sealed class FhAudio {
```

fkelava (changes requested):

src/core/ffx/ids/sound.cs
DawnVeilwinter: I'm not convinced it looks any better that way. If we discover these are actually signed integers where the sign bit is used for tagging, we could change `0x80000001` to `-1` though. FMOD Event Viewer — the tool normally used to browse FMOD event files — shows the sound numbers as decimal too, so that way the IDs would match up a little better, even if the representation may feel a bit jarring at first.

fkelava (approved):

PR #306, opened 2026-09-16
Remove S1 EH; turn S0 into a stub native debugger

fkelava:
In the EH discussions in `a11` (ex. #169) we discussed that the implementation given then (in-process {S|V}EH) was merely a stopover until we could do something better. This is the first step in that.

`MiniDumpWriteDump` documentation calls out the better approach: (https://learn.microsoft.com/en-us/windows/win32/api/minidumpapiset/nf-minidumpapiset-minidumpwritedump#remarks)
> MiniDumpWriteDump should be called from a separate process if at all possible, rather than from within the target process being dumped. This is especially true when the target process is already not stable. For example, if it just crashed.

Most interpret this as a crash handler that can be invoked on fault by a in-process EH function.

But we deviate a bit from this design; we already _have_ an external process to leverage, the Stage 0 loader. Instead of the in-process crash handler calling through, we could make it into a Win32 native debugger, and respond to exception events. This is what this PR does, complete with native stack walking and symbol support.

Next to be implemented is mixed-mode stack walking, so the managed frames on the stack can also be backtraced.


src/stage0/src/main.cpp
```diff
@@ -3,45 +3,62 @@
...
+ */
+
 #include "fhstage0.h"
 
-int wmain(int argc, wchar_t* argv[ ]) {
+void stage0_dbg_loop(); // Forward declaration of debugger loop function.
+
+int wmain(
+    int      argc,
+    wchar_t* argv[ ]
```
DawnVeilwinter: Stylistically, I don't think the space in `argv[ ]` makes much sense here?

src/stage0/src/main.cpp
```diff
@@ -51,60 +68,43 @@ int wmain(int argc, wchar_t* argv[ ]) {
...
-    // STEP 6:
-    // Wait for program to (un)naturally terminate.
-    //
+    // Either wait for the process to exit if an external debugger is connected,
+    // or begin pumping debug events with the Stage 0 stub debugger.
+    if (external_debug) {
+        ResumeThread       (pi.hThread);
+        WaitForSingleObject(pi.hProcess, INFINITE);
+    }
+    else { stage0_dbg_loop(); }
```
DawnVeilwinter: If the `if` is multiline, I think the `else` should be as well.

src/stage0/src/main.cpp
```diff
@@ -3,45 +3,62 @@
...
+    wchar_t* argv[ ]
+) {
     if (argc < 2) {
         std::wcerr << "Invalid call. You must specify an executable to launch.\n";
         std::wcerr << "Usage: fhstage0.exe {EXECUTABLE_TO_LAUNCH} {ARGS}\n";
         return 1;
     }
 
-    LPCSTR              szDllPath  = "fhstage1.dll";
+    LPCSTR              path_dll = "fhstage1.dll";
```
DawnVeilwinter: We usually use the `generic_specific` guideline when there are multiple identifiers with the same `generic`. This is not the case here, and I would prefer the reading order of `dll_path`.

src/stage1/src/main.cpp
```diff
@@ -67,25 +67,25 @@ static bool load_hostfxr() {
 // Runs before the program's own entrypoint, setting up Fahrenheit.
 static int stage1_main(void) {
     // STEP 5:
-    // If supported, install an EH override which allows us to capture
-    // a customized core dump for easier debugging.
+    // If necessary, suppress the game's SEH filters, so Stage 0
```
DawnVeilwinter, line 85: You say "If necessary," but then do it unconditionally?

src/stage1/src/eh.cpp
```diff
@@ -165,28 +36,17 @@ BOOL stage1_eh_install(LPBYTE ptr_main_module) {
...
 
-    ::SetThreadDescription(g_eh_thread_handler, L"Fahrenheit EH");
-    SetUnhandledExceptionFilter(&stage1_eh);
+    // We don't care about the original SEH filter in the slightest, so we don't keep it.
+    void* fnptr_eh_original = NULL;
 
-    if (MH_CreateHookApi(L"kernel32.dll", "SetUnhandledExceptionFilter", &stage1_eh_set_filter, reinterpret_cast<void**>(&g_fnptr_eh_original)) != MH_OK
-    ||  MH_EnableHook   (&SetUnhandledExceptionFilter)                                                                                          != MH_OK) {
-        std::wcerr << "Failed to install EH hook for " << exe_name << std::endl;
+    if (MH_CreateHookApi(L"kernel32.dll", "SetUnhandledExceptionFilter", &stage1_eh_set_filter, &fnptr_eh_original) != MH_OK
```
DawnVeilwinter: I think you meant to pass in `NULL` for the `ppOriginal` argument?
```suggestion
    // We don't care about the original SEH filter in the slightest, so we don't keep it.
    if (MH_CreateHookAPI(L"kernel32.dll", "SetUnhandledExceptionFilter", &stage1_eh_set_filter, NULL) != MH_OK
```

src/stage1/src/eh.cpp
```diff
@@ -165,28 +36,17 @@ BOOL stage1_eh_install(LPBYTE ptr_main_module) {
...
+    ||  MH_EnableHook   (&SetUnhandledExceptionFilter)                                                              != MH_OK) {
```
DawnVeilwinter: Optionally, I think this alignment is a bit much.
```suggestion
    ||  MH_EnableHook   (&SetUnhandledExceptionFilter) != MH_OK) {
```

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,608 @@
...
+    HANDLE   h_process,
+    HANDLE   h_thread,
+    CONTEXT* ptr_context
+) {
+    /* [fkelava 13/09/26 13:49]
+     * https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-stackwalk64
+     * > This context may be modified, so do not pass a context record that should not be modified.
+     *
+     * We _first_ dump core precisely so the context can be freely modified by the stack walker.
+     */
```
DawnVeilwinter: I believe this comment should be moved to `stage0_dbg_exception`, above the call to `stage0_dbg_create_dump`.

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,608 @@
...
+    if (sz_module_path > MAX_PATH) {
+        std::wcerr << "[!] GetFinalPathNameByHandleW() - path length exceeded MAX_PATH" << std::endl;
+        error_code = ERROR_BUFFER_OVERFLOW;
+
+        return FALSE;
+    }
+
+    /* [fkelava 13/09/26 16:43]
+     * When deferred symbols are in use, the correct DLL size must be passed.
+     * See https://groups.google.com/forum/#!topic/comp.os.ms-windows.programmer.win32/ulkwYhM3020
```
DawnVeilwinter: This links gives me a `429 Too Many Requests` error. A brief summary may be appreciated? Optional.

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,608 @@
...
+        error_code = GetLastError();
+
+        return FALSE;
+    }
+
+#if _DEBUG
+    std::wcout << "Module loaded: " << module_path << std::endl;
+#endif
+    /* [fkelava 13/09/26 02:03]
+     * > The debugger should close the handle to the DLL while processing LOAD_DLL_DEBUG_EVENT.
```
DawnVeilwinter, line 532: Where is this quote from?
Edit: I feel like you moved it here from `stage0_dbg_loop`, not realizing the reader of this function does not have the context of unattributed segments being from [Debugging Events](https://learn.microsoft.com/en-us/windows/win32/debug/debugging-events) / [Writing the Debugger's Main Loop](https://learn.microsoft.com/en-us/windows/win32/debug/writing-the-debugger-s-main-loop).

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,608 @@
...
+        std::wcerr << "[!] GetModuleFileNameW() failed, error code: " << GetLastError() << std::endl;
+        return FALSE;
+    }
+
+    /* [fkelava 15/09/26 14:54]
+     * We have to remove the last path element twice to get from /bin/fhstage0.exe to the base directory.
+     */
+    if (PathCchRemoveFileSpec(path_dir_base, MAX_PATH) != S_OK ||
+        PathCchRemoveFileSpec(path_dir_base, MAX_PATH) != S_OK) {
+        std::wcerr << "[!] PathCchRemoveFileSpec() failed" << std::endl;
```
DawnVeilwinter: It would be nice to print the path in this case.

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,608 @@
...
+    }
+
+    return TRUE;
+}
+
+// The main loop of the debugger. Handles incoming debug events.
+void stage0_dbg_loop() {
+    BOOL init_failed = FALSE;
+    if (!stage0_dbg_init()) {
+        std::wcerr << "Failed to create cache and crash directories. Aborting." << std::endl;
```
DawnVeilwinter: I believe `"Failed to create debugger directories. Aborting."` would suffice.

DawnVeilwinter (changes requested):
Great work! It's incredible how much you managed despite the lack of good learning resources.

src/stage0/src/dbg.cpp
fkelava: That is a verbatim one-liner from that thread. No elaboration is given, though I have confirmed it to be true.

src/stage1/src/main.cpp
fkelava: That function aborts if it doesn't recognize the target binary. This is handy for those porting Fahrenheit to third-party titles.

DawnVeilwinter (dismissed):
Looks good so far!

src/stage0/src/dbg.cpp
DawnVeilwinter: I believe in this case we should use the style used in other places in the file:
```
<link>
> Quote
```

src/stage0/src/main.cpp
```diff
@@ -22,49 +22,168 @@
...
+}
+
 int wmain(
     int      argc,
     wchar_t* argv[]
 ) {
     if (argc < 2) {
         fwprintf_s(stderr, L"Invalid call. You must specify an executable to launch.\n");
-        fwprintf_s(stderr, L"Usage: fhstage0.exe {EXECUTABLE_TO_LAUNCH} {ARGS}\n");
+        fwprintf_s(stderr, L"Usage: fhstage0.exe {TARGET} {ARGS_STAGE0} -- {ARGS_TARGET}\n");
```
DawnVeilwinter: We should include information about valid arguments. Also, I believe it is standard to use square brackets to denote placeholders?
```suggestion
        fwprintf_s(stderr, L"Usage: fhstage0.exe [target] [options] -- [target-options]\n");
        fwprintf_s(stderr, L"Options:\n");
        fwprintf_s(stderr, L"  --debug    Allow for the use of an external debugger.\n");
```

src/stage0/src/main.cpp
```diff
@@ -106,7 +267,7 @@ int wmain(
 
     if (exit_code != 0) {
         fwprintf_s(stdout, L"Process exited with code 0x%X.\n", exit_code);
-        fwprintf_s(stdout, L"If reporting an issue, please include any core dump (*.dmp) you see in the game directory.\n");
+        fwprintf_s(stdout, L"If reporting an issue, include the core dump (*.dmp) mentioned above in the output.\n");
```
DawnVeilwinter: The "please" was fine, I think. There are valid concerns (PII etc.) that may stop someone from sharing a core dump, after all.
```suggestion
        fwprintf_s(stdout, L"If reporting an issue, please include the core dump (*.dmp) from the location mentioned above."
```

DawnVeilwinter (changes requested):

DawnVeilwinter (dismissed):

DawnVeilwinter (dismissed):
Looks fine. Will need a final look-through once it's no longer a draft.

.editorconfig
```diff
@@ -1,12 +1,12 @@
-﻿[*.cs]
+﻿# indentation style
+indent_style = space
+indent_size = 4
+
+[*.cs]
 
 # guidelines
 guidelines = 100
```
DawnVeilwinter: I think the guideline should also be generic, if possible.
```suggestion
guidelines = 100

[*.cs]
```

src/stage1/inc/nethost/README.md
```diff
@@ -0,0 +1,23 @@
...
+in which they suppress building the `.vcxproj` normally and drive the build from a .NET project target.
+
+"Relatively complicated" is, put charitably, an understatement. We must co-opt the separate .NET project,
+but use it merely to copy the headers and libraries from your local .NET SDK install to the right places. See `src/stage1/Fahrenheit.Stage1.Deps.csproj`.
+
+[!WARNING]
+> Do not commit the copied-over headers or libraries to this repository.
+
+See generally https://github.com/dotnet/samples/blob/86ff8487361a6f32549d9c9ab8b14dde55c643cf/core/hosting/readme.md,
+https://github.com/dotnet/samples/blob/86ff8487361a6f32549d9c9ab8b14dde55c643cf/core/hosting/src/NativeHost/NativeHost.csproj.
```
DawnVeilwinter: Since this is a markdown file, let's use link masking to make this easier to read when rendered.
```suggestion
See generally [the .NET Hosting Sample](https://github.com/dotnet/samples/blob/86ff8487361a6f32549d9c9ab8b14dde55c643cf/core/hosting/readme.md)
and [its project file](https://github.com/dotnet/samples/blob/86ff8487361a6f32549d9c9ab8b14dde55c643cf/core/hosting/src/NativeHost/NativeHost.csproj).
```

DawnVeilwinter (changes requested):
Just some minor nitpicks!

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,1127 @@
...
+
+wchar_t g_path_coreclr     [MAX_PATH] = { 0 }; // The full path to the loaded CoreCLR.
+wchar_t g_path_mscordbi    [MAX_PATH] = { 0 }; // The full path to the `mscordbi` module for the given CoreCLR.
+wchar_t g_path_mscordacwks [MAX_PATH] = { 0 }; // The full path to the `mscordacwks` module for the given CoreCLR.
+wchar_t g_path_mscordaccore[MAX_PATH] = { 0 }; // The full path to the `mscordaccore` module for the given CoreCLR.
+
+std::map   <std::wstring, bool> g_map_checked_symbol; // Whether we performed symbol file lookup for a given module.
+std::vector<std::wstring>       g_frames_managed;     // A list of managed frame strings. Used to fill the gaps in the native stack walk.
+std::vector<std::wstring>       g_frames_native;      // A list of native frame strings.
+std::vector<S0_FRAME_TYPE>      g_frames_type;        // A list of frames, indicating the type of any given frame.
```
DawnVeilwinter: These should probably be combined, unless there is something precluding us from doing so?
```cpp
struct FrameData {
    std::wstring  string;
    S0_FRAME_TYPE type;
};

std::vector<FrameData> g_frames;
```

fkelava: I tried and found it too difficult to reconcile two separate stack walks that can fail at any point, for any reason.

src/stage0/src/main.cpp
```diff
@@ -173,21 +232,25 @@ int wmain(
         return 1;
     }
 
-    HRESULT hr;
+    if (!s0_main_init()) {
+        fwprintf_s(stderr, L"Stage 1 failed to initialize.\n");
```
DawnVeilwinter: Did you mean:
```suggestion
        fwprintf_s(stderr, L"Stage 0 failed to initialize.\n");
```

src/stage0/src/dbg.cpp
```diff
@@ -17,11 +17,461 @@
...
+
+wchar_t g_path_dir_cache[MAX_PATH] = { 0 }; // The full path to the 'cache' directory, used to store symbols.
+wchar_t g_path_dir_crash[MAX_PATH] = { 0 }; // The full path to the 'crash' directory, used to store core dumps.
+
+wchar_t g_path_coreclr     [MAX_PATH] = { 0 }; // The full path to the loaded CoreCLR.
+wchar_t g_path_mscordbi    [MAX_PATH] = { 0 }; // The full path to the `mscordbi` module for the given CoreCLR.
+wchar_t g_path_mscordacwks [MAX_PATH] = { 0 }; // The full path to the `mscordacwks` module for the given CoreCLR.
+wchar_t g_path_mscordaccore[MAX_PATH] = { 0 }; // The full path to the `mscordaccore` module for the given CoreCLR.
+
+std::map   <std::wstring, bool> g_map_checked_symbol; // Whether we performed symbol file lookup for a given module.
```
DawnVeilwinter: Since all we're storing is booleans, this is probably better as an [unordered set](https://en.cppreference.com/cpp/container/unordered_set)?
Any code checking the boolean should be replaced with [`unordered_set::contains`](https://en.cppreference.com/cpp/container/unordered_set/contains).

src/stage0/src/dbg.cpp
```diff
@@ -17,11 +17,461 @@
...
+LPVOID g_ptr_coreclr; // The pointer to `coreclr.dll` in memory.
+
+/* [fkelava 19/09/26 00:50]
+ * Here we simultaneously borrow a bit and yet diverge from Dalamud.
+ * The different choice of interface seems more cosmetic than anything, though.
+ *
+ * https://github.com/goatcorp/Dalamud/blob/e81744f6aea94bb6781affdd0d0b9319592f95d9/DalamudCrashHandler/DalamudCrashHandler.cpp#L306
+ *
+ * Due to the almost _nonexistent_ documentation, any difference in approach is to be taken
+ * as the unfortunate result of a goat and a donkey having to stumble around in the dark.
```
DawnVeilwinter, line 48: ??? okay?

src/stage0/src/dbg.cpp
```diff
@@ -158,9 +632,48 @@ static void stage0_dbg_print_context(
...
+            case FRAME_UNKNOWN:
+                frame = L"Unknown frame.\n";
+                break;
+
+            case FRAME_MANAGED:
+                /* [fkelava 21/09/26 16:30]
+                 * We might not have data for all managed frames.
+                 * Currently, we don't handle things like forward and reverse P/Invoke stubs.
+                 */
+                frame = count_managed >= g_frames_managed.size()
```
DawnVeilwinter, line 655: Will this simple method of determining what frame are managed here not fail when the code goes from managed to native more than once? (e.g. calling a game function with a function pointer to managed code)

DawnVeilwinter (changes requested):

DawnVeilwinter:
I'm converting this back to a draft while you work out the managed debugger stuffs

src/core/bootloader.cs
```diff
@@ -145,11 +145,11 @@ internal sealed class FhLoader {
     private readonly Dictionary<string, FhLoadContext> _load_contexts = [];
 
     internal FhLoader() {
-        // Loading the core library into ALC.Default ensures it does not 'leak' into plugins' load contexts, causing type identity mismatches.
-        Assembly self = AssemblyLoadContext.Default.LoadFromAssemblyPath(Path.Join(FhEnvironment.Finder.Binaries.FullName, "fh.dll"));
+        // The core library is loaded into ALC.Default. This ensures it does not 'leak' into plugins' load contexts, causing type identity mismatches.
```
DawnVeilwinter: Fahrenheit uses the term "mods", not "plugins".

src/stage0/README.md
```diff
@@ -1,11 +1,12 @@
...
 In order, it:
 - Creates the target process in ``CREATE_SUSPENDED`` state.
 - Modifies the process' PE header and IAT to:
     - Add the Fahrenheit "Stage 1" Loader to the head of the import list, so it executes first.
 	- This process is performed using MS Detours' [``DetourCreateProcessWithDll``](https://github.com/microsoft/detours/wiki/DetourCreateProcessWithDll).
-- Captures console output from the game.
-- Waits for the game process to exit, capturing its exit code.
+- Captures standard I/O from the game.
+- Waits for the game process to crash or exit, capturing its exit code.
+    - If the game process crashed, it prints exception information, creates a core dump, and displays an approximate stack trace.
```
DawnVeilwinter: We don't need to specify the stack trace is "approximate". Also, being obtained with use of hueristics does not make something approximate.

src/stage1/src/eh.cpp
```diff
@@ -3,190 +3,67 @@
...
-
-    return TRUE;
+static LPTOP_LEVEL_EXCEPTION_FILTER WINAPI stage1_eh_set_filter(LPTOP_LEVEL_EXCEPTION_FILTER fnptr_exception_filter) {
+    return NULL;
 }
 
-/*
- * Writes a customized core dump.
- */
+BOOL stage1_eh_suppress() {
```
DawnVeilwinter: As with `stage0 > s0`, this should be named `s1_eh_suppress()`.

src/stage1/src/eh.cpp
```diff
@@ -3,190 +3,67 @@
...
-            return FALSE;
-
-        case IncludeThreadCallback: {
-            // Exclude the thread which writes the minidump.
-            return ptr_callback_input->IncludeThread.ThreadId != g_eh_thread_handler_id;
-        } break;
-    }
-
-    return TRUE;
+static LPTOP_LEVEL_EXCEPTION_FILTER WINAPI stage1_eh_set_filter(LPTOP_LEVEL_EXCEPTION_FILTER fnptr_exception_filter) {
```
DawnVeilwinter: As with `stage0 > s0`, this should be named `s1_eh_set_filter`.

src/stage1/src/main.cpp
```diff
@@ -31,32 +30,42 @@ hostfxr_close_fn                         g_fnptr_hostfxr_close;
 FILE* g_stdout;
 FILE* g_stderr;
 
-BOOL stage1_eh_install(LPBYTE ptr_main_module); // Forward declaration of EH installer function
+BOOL stage1_eh_suppress(); // Forward declaration of EH suppressor function
```
DawnVeilwinter: As with `stage0 > s0`, this should be named `s1_eh_suppress`.

src/stage1/src/main.cpp
```diff
@@ -31,32 +30,42 @@ hostfxr_close_fn                         g_fnptr_hostfxr_close;
...
 
-/*
- * Uses the `nethost` library to discover the location of the .NET hosting library,
- * `hostfxr`, and obtains the necessary function pointers from it.
- */
+// Uses the `nethost` library to discover the location of the .NET hosting library,
+// `hostfxr`, and obtains the necessary function pointers from it.
+static BOOL stage1_load_hostfxr(
```
DawnVeilwinter: As with `stage0 > s0`, this should be named `s1_load_hostfxr`.

src/stage1/src/main.cpp
```diff
@@ -66,74 +75,78 @@ static bool load_hostfxr() {
 
 // Runs before the program's own entrypoint, setting up Fahrenheit.
 static int stage1_main(void) {
```
DawnVeilwinter: As with `stage0 > s0`, this should be named `s1_main`.

src/stage1/src/main.cpp
```diff
@@ -176,99 +188,67 @@ static int stage1_main(void) {
...
- * 
- * This is required because certain other tools expect the game's working directory to be
- * unmodified when they load into the process, which occurs immediately after Stage 1 exits `DllMain`.
- */
-static BOOL stage1_init() {
-    // STEP 2:
-    // Attach to the Stage0 console and forward stdout/stderr to it.
+
+// Records the directory we started from and hooks the target's entrypoint.
+static BOOL stage1_init(
```
DawnVeilwinter: As with `stage0 > s0`, this should be named `s1_init`.

DawnVeilwinter (changes requested):
I've reviewed everything *except* `stage0/src/dbg.cpp`. I will review the rest shortly.

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,1128 @@
...
+
+    wchar_t sym_native[MAX_SYM_NAME] = { 0 };
+    DWORD64 sym_displacement = 0;
+
+    if (!SymFromAddrW(h_process, frame_addr, &sym_displacement, &sym.si)) {
+        /* [fkelava 21/09/26 14:53]
+         * We may not have the PDB or any other symbols for the target binary.
+         * In this case SymFromAddrW seems to return ERROR_INVALID_ADDRESS.
+         *
+         * But, e.g., FFX+193912 is suitable and useful, so we display that instead.
```
DawnVeilwinter: This should probably contain the actual format: `FFX.exe+0x19A310`.

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,1128 @@
...
+    }
+    else {
+        swprintf_s(sym_native, MAX_SYM_NAME, L"%s!%s+0x%llX\n", module.ModuleName, sym.si.Name, sym_displacement);
+    }
+
+    g_frames_native.emplace_back(sym_native);
+    return FRAME_NATIVE;
+}
+
+// Prints the register state at the time an exception was caught.
```
DawnVeilwinter: This may be a reach, but could we also print the FPU stack?

src/stage0/src/dbg.cpp
```diff
@@ -0,0 +1,1128 @@
...
+      | MiniDumpWithFullMemoryInfo
+      | MiniDumpWithThreadInfo
+      | MiniDumpWithProcessThreadData
+      | MiniDumpWithUnloadedModules);
+
+    /* [fkelava 11/06/26 21:24]
+     * MiniDumpWriteDump expects, in MINIDUMP_EXCEPTION_INFORMATION, a PEXCEPTION_POINTERS
+     * (a CONTEXT and EXCEPTION_RECORD). EXCEPTION_DEBUG_INFO only gets the latter.
+     *
+     * GetThreadContext solves that, but there's a catch. MINIDUMP_EXCEPTION_INFORMATION has a ClientPointers field:
```
DawnVeilwinter: "has a ClientPointers field **that**:"

DawnVeilwinter (changes requested):
Smol documentation nitpicks only!

src/stage0/src/dbg.cpp
fkelava: The actual format in a trace would be `FFX+0x{OFFS}.` Amended accordingly.

DawnVeilwinter (approved):
Looks good to me, thank you!

PR #309, opened 2026-09-18
Fix some sloppy zeroing and null-termination

fkelava:
For a long time, we knew this game didn't like running specifically under Fahrenheit combined with a native debugger. While checking why, it was unraveled that this had nothing to do with a native debugger per se. Rather, for reasons yet (and probably forever) unknown, this particular environment unraveled our invalid and/or careless behavior.

Per the C# language specification for `stackalloc` (https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc):
> The content of the newly allocated memory is undefined. You should initialize it before it's used, either with a stackalloc initializer or a method like [Span<T>.Clear](https://learn.microsoft.com/en-us/dotnet/api/system.span-1.clear).

Also, `u8` literals are not null-terminated, but we were passing them to functions that expect null-terminated strings.

Fixing this restores the capacity to properly boot under pure native debuggers- a prerequisite for #306.


DawnVeilwinter (approved):
I believe the Blake3 hash buffer clears to be unnecessary, but it doesn't hurt either.
Merge at will.

PR #311, opened 2026-09-19
Set version references to `alpha12`

fkelava:
Now that `alpha11` has been published on NuGet, we need to switch the local version refs to the next release so local SDKs do not conflict with the official `alpha11` SDK.


EvelynTSMG (approved):
Admittedly I still find this workflow of bumping the version early weird, but I see no errors in the reasoning.

DawnVeilwinter (approved):
Admittedly I still find this workflow of bumping the version early weird, but I see no errors in the reasoning.

